PR #157, opened 2024-12-24
updated German translation

InvestigateXM:
updated German translations to use official terms used in Elite in German as well as fixing some translation errors due to missing context

only translated the strings that are present in the dedicated file for them (/Properties), as I don't wanna break the rest of the program if I start digging around for the other strings

### DIFF
--- a/SrvSurvey/Properties/Misc.de.resx
+++ b/SrvSurvey/Properties/Misc.de.resx
@@ -84,11 +84,11 @@
     <comment>⏷ Show list</comment>
   </data>
   <data name="FormBoxelSearch_SpanshUpdated">
-    <value>Spanisch aktualisiert: {0}</value>
+    <value>Spansh aktualisiert: {0}</value>
     <comment>Spansh updated: {0}</comment>
   </data>
   <data name="FormBoxelSearch_SpanshUpdatedBodies">
-    <value>Spanisch aktualisiert: {0}, Körper: {1}</value>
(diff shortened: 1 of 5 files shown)


SrvSurvey/Properties/Misc.de.resx
```diff
@@ -243,7 +243,7 @@ Möchten Sie es erstellen?</value>
     <comment>Cannot find folder:
 {0}
 
-Would you like to create it?</comment>
+Möchtest du ihn erstellen?</comment>
```
njthomson, line 246: The comment needs to be the original English phrase - this is how I tell when new translations are required.

SrvSurvey/Properties/Misc.de.resx
```diff
@@ -173,10 +173,10 @@
   </data>
   <data name="PlotGalMap_Destination">
     <value>Bestimmungsort:</value>
-    <comment>Destination:</comment>
+    <comment>Ziel:</comment>
```
njthomson, line 176: The `comment` line should remain in English, but should the `value` line above change?

njthomson (changes requested):
Danke @InvestigateXM for these updates! There are 2 places I noticed where you translated the comment rather than the original phrase. These need to remain as the original English phrase - this is how I tell when new translations are required.

InvestigateXM:
My mistake @njthomson, thats what I get for doing stuff during the night. Newest commit should have fixed these issues :)

njthomson (approved):

PR #179, opened 2025-02-09
Spanish Localization

Transcan:
All available translations done.
Maybe I localized some "in-work" files (like te codex one)... Sorry, I got excited and translated everything I saw. ^_^;
I will translate as you enable new panels.


njthomson:
Nice work @Transcan - you'll see these in the next release build.

PR #180, opened 2025-02-09
Add reminder for refuel systems on Spansh routes

Eearslya:
Adds a small reminder note when following Spansh routes if the next system is meant to be used for refueling.

image


SrvSurvey/plotters/PlotSphericalSearch.cs
```diff
@@ -387,6 +387,12 @@ private void drawRouteFollow(bool drawLine)
                 newLine(true);
```
njthomson, line 387: I'm curious - does it not display anything about refueling from the notes? (That was the design for that string field)

Eearslya: It definitely didn't display anything before I made this change, I didn't realize it was meant to. The notes do appear in the route form, but they're null in the `FollowRoute` object.

image
image

njthomson: Take a look at the file in `%appdata%\SrvSurvey\SrvSurvey\1.1.0.0\routes\` - the notes should be stored there.

njthomson: These notes should also appear in the FSD jump overlay when jumping to that system.

Eearslya: The notes are unfortunately not in that file, nor did they appear on the FSDJump overlay. Somehow, they live only in that import window.

```json
{
  "active": false,
  "autoCopy": true,
  "last": 173,
  "hops": [
    {"name":"Sagittarius A*","id64":20578934,"x":25.21875,"y":-20.90625,"z":25899.96875},
    {"name":"Juenae CE-C c2-6147","id64":1689758413563666,"x":19.34375,"y":-85.375,"z":25906.875,"refuel":true},
    {"name":"Juenae HF-A c3-2295","id64":630928648873754,"x":4.65625,"y":-139.40625,"z":25940.15625,"refuel":true},
    {"name":"Juenae CX-A d1-1498","id64":51481365418891,"x":-41.9375,"y":-160.5625,"z":25974.3125},
    {"name":"Phroi Pri VD-Z d1-7664","id64":263343495418771,"x":-99.90625,"y":-171.40625,"z":26002.03125,"refuel":true},
    {"name":"Phroi Pri TN-W c4-2576","id64":708169072256810,"x":-154.59375,"y":-179.65625,"z":26033.6563,"refuel":true},
 ...
 ```

njthomson: Was this a galaxy plotter route? Trying again locally, I see that has no notes saved. Hmm...

njthomson: Ah, I see - those never were part of notes but shown in the UI. https://github.com/njthomson/SrvSurvey/blob/main/SrvSurvey/forms/FormRoute.cs#L269

So your change is a good thing. Only I am surprised the next text is orange (and not following the theme colour).

Eearslya: True, I thought orange would work because it's the color I think of when I think of fuel scooping, but I should probably be testing things like this with the default theme.

Question though, since you said that the whole purpose of the `notes` field was for exactly things like this, would it be better to adjust things so that those notes do get put in there, rather than a new line?

njthomson: I've been pondering either making the data more structured, putting scan targets in the .json in some form, vs lumping refuel and neutrons into notes. Given the wave of interest to get SrvSurvey localized it would make sense to remove a general "notes" field and store data in a more structured way. It's a bit more work but it's the cleaner way to engineer things.

Eearslya:
Alright, I've re-done this PR to include both Refuel and Neutron "alerts". I'm still not too familiar with C# so I didn't know how to put these into the resource file for localization, but it's at least functional.

image

image

njthomson:
> Alright, I've re-done this PR to include both Refuel and Neutron "alerts". I'm still not too familiar with C# so I didn't know how to put these into the resource file for localization, but it's at least functional.
> 
> image
> 
> image

If you're interested, I can tell you how to do the string resources, it's not that hard, but I'm also very willing to do that afterwards if you prefer.

charxon:
Thanks all for your hard work. I would love to use this feature. Is there any progress in it? If there is something I can help, I can offer my limited knowledge :)

PR #183, opened 2025-02-13
Remove new discovery flags upon codex scan

Eearslya:
SRV Survey will display a solid or hollow flag and color a biological signal gold if it is the commander's first time seeing it, or the first time seeing it for a region. This flag and coloration remains there, even after the commander has scanned the biological signal and added it to their codex.

This update would change it so that the biological signal returns to normal colors and removes the flag icon once the signal has been scanned. This will allow players to more easily keep track of exactly what they still need to discover within a system.

Before scans:
image

After fully scanning Fonticulua and partially scanning Bacterium:
image
image

This does remove a bit of the "caching" functionality that was present in the `isCmdrFirst` and `isNewEntry` properties. I'm not sure if this was a performance optimization or not, so I'm open to suggestions if you think these changes might cause problems.

### DIFF
--- a/SrvSurvey/plotters/PlotBioSystem.cs
+++ b/SrvSurvey/plotters/PlotBioSystem.cs
@@ -215,8 +215,8 @@ namespace SrvSurvey.plotters
 
                     // do we already know if this is a first discovery?
                     string? discoveryPrefix = null;
-                    if (organism.isCmdrFirst) discoveryPrefix = "⚑";
-                    else if (organism.isNewEntry) discoveryPrefix = "⚐";
+                    //if (organism.isCmdrFirst) discoveryPrefix = "⚑";
+                    //else if (organism.isNewEntry) discoveryPrefix = "⚐";
 
                     dty = (int)dty;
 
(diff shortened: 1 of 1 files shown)


njthomson:
@Eearslya it was an intentional design that the flags remain visible and gold, mostly as eye candy to remind cmdr's that they found something new in this system.

Personally I found the excess of hollow yellow flags to be annoying, especially when entering a new galactic region, but I enjoy seeing solid gold flags remain. There is setting `Highlight ⚐ regional firsts in gold` to make only the solid flags be gold, though I happen like that they remain. Still ... I recognize it's completely subjective what is "satisfying eye-candy" vs annoying clutter.

If you fancy it ... may I recommend adding a new setting, something like "Remove ⚑ and ⚐ flags once scanned" - so that your new code fires when it is active, or the previous behaviour remains if not. That existing setting is named `highlightRegionalFirsts` - take a look where that is used and try something a similar?

Roughly it would mean:
- Add an entry in settings.cs with the default value assigned.
- Add another CheckBox in FormSettings.cs, make sure the `Tag` value is the name of the setting.
- Add code/IF's in PlotBioSystem using `Game.settings.<new name>` to light up your changes or keep the old behaviour.

I should add: the UI in FormSettings is quite cluttered, but you could shift the two Clear buttons up a bit to make room ...
image
... make sure to leave some horizontal space between the Checkbox's otherwise they will overlap when large-font/high-dpi scaling happens.

And make sure your system has no (100%) scaling when editing the forms - otherwise the Visual Studio kinda applies the scaling in reverse causing all sorts of layout troubles.

PR #188, opened 2025-02-15
Updated Spanish human translation

Transcan:


njthomson:
Nice work @Transcan !

PR #190, opened 2025-02-16
Simplified Chinese localization

ckx000:
add Simplified Chinese localization

I am following the format of other languages in the code files to proceed, and I have added an entry in Localize.cs (I guess this is where language entries are added).

Since I haven't checked the results through compilation.I'm not entirely sure if the Chinese font will display correctly in the overlays


njthomson:
> add Simplified Chinese localization
> 
> I am following the format of other languages in the code files to proceed, and I have added an entry in Localize.cs (I guess this is where language entries are added).
> 
> Since I haven't checked the results through compilation.I'm not entirely sure if the Chinese font will display correctly in the overlays

I will try this out tomorrow and let you know what I find.

njthomson:
> add Simplified Chinese localization
> 
> I am following the format of other languages in the code files to proceed, and I have added an entry in Localize.cs (I guess this is where language entries are added).
> 
> Since I haven't checked the results through compilation.I'm not entirely sure if the Chinese font will display correctly in the overlays

Looks like there are a few files with mismatched XML tags:

image

njthomson:
@ckx000 - whilst it will cause you to have to reconcile many changes (and I apologise for that), I would like to enable simplified Chinese with machine translations first, and have your PR go after - this way we can track what you have translated already vs the new resources that have been recently added.

ckx000:
> > add Simplified Chinese localization
> > I am following the format of other languages in the code files to proceed, and I have added an entry in Localize.cs (I guess this is where language entries are added).
> > Since I haven't checked the results through compilation.I'm not entirely sure if the Chinese font will display correctly in the overlays
> 
> Looks like there are a few files with mismatched XML tags:
> 
> image
> 
> Outside of those, things looks kinda okay - though perhaps I need to add resources so those material names could be translated?
> 
> image
> 
> image

Indeed, there is still a lot to be done, and many areas need to add multilingual support.
I want to first try using a bilingual approach for some proprietary material names in the game to see how it looks, hoping it won't exceed the display limits.

ckx000:
> @ckx000 - whilst it will cause you to have to reconcile many changes (and I apologise for that), I would like to enable simplified Chinese with machine translations first, and have your PR go after - this way we can track what you have translated already vs the new resources that have been recently added.
> 
> For more details about recent updates to my localization processes, see: https://github.com/njthomson/SrvSurvey/blob/main/SrvSurvey/Properties/_loc_readme.md

Of course. In fact, this should be the best approach... 
You update the code first, and then I will submit the updates afterward.

Shall we close this pull then?

njthomson:
> > > add Simplified Chinese localization
> > > I am following the format of other languages in the code files to proceed, and I have added an entry in Localize.cs (I guess this is where language entries are added).
> > > Since I haven't checked the results through compilation.I'm not entirely sure if the Chinese font will display correctly in the overlays
> > 
> > 
> > Looks like there are a few files with mismatched XML tags:
> > image
> > Outside of those, things looks kinda okay - though perhaps I need to add resources so those material names could be translated?
> > image
> > image
> 
> Indeed, there is still a lot to be done, and many areas need to add multilingual support. I want to first try using a bilingual approach for some proprietary material names in the game to see how it looks, hoping it won't exceed the display limits.

My memory of localizing things at work is that Chinese tends to be shorter, though obviously we need to account for the widths of all languages.

I'm slowing plugging away at getting all the windows and overlays localizable, it's a quite a bit of work that will take some time to complete.

ckx000:
> 
> My memory of localizing things at work is that Chinese tends to be shorter, though obviously we need to account for the widths of all languages.
> 
> I'm slowing plugging away at getting all the windows and overlays localizable, it's a quite a bit of work that will take some time to complete.

Normally, that's correct..... It's just that in some places I want to use a display format like "Sinuous Tubers(蜿蜒块茎)",

Because currently, there isn't a fixed Chinese term among Chinese players for mutual communication, 
I hope to use a bilingual display to provide a transition for players using SrvSurvey.

ckx000:
I have redone it using your new template.

ckx000:
The code structure and version differences were messed up by "GitHub Desktop," or it might be because I'm not very skilled at using this tool.

In the end, I manually replaced and fixed them one by one.

ckx000 (commented):
checked

njthomson:
@ckx000 - these changes look good.

I must apologise again as I shifted some resources from `misc.resx` over to `PlotSysStatus.resx` and `PlotGalMap.resx` - however they are unchanged, so you should be able to copy & paste what you have into those new files. And strip the `PlotSysStatus_`/PlotGalMap_` prefix from the resource names too.

ckx000:
Oh damn, made a mistake... Accidentally synced the branch, should have synced my main branch first..

PR #347, opened 2025-08-04
Updata Simplified Chinese localization

ckx000:
- Update translations

- Restore the Commodities original text for easier user cross-reference with in-game content.


njthomson (approved):
Nice!

PR #371, opened 2025-08-19
Update French localization, sort commodities and their categories based on their translation

Entrivax:
- Updated some of the French labels (and verified some others)
- Added a `CommodityCategories.fr.resx` file to translate commodities categories to French
- Added a sorting in `PlotBuildCommodities.cs` of the categories and the commodities based on their translation value so it can match the in-game market.

### DIFF
--- a/SrvSurvey/forms/FormGroundTarget.fr.resx
+++ b/SrvSurvey/forms/FormGroundTarget.fr.resx
@@ -485,12 +485,12 @@
   </data>
   <data name="btnTargetCurrent.Text">
     <!--Machine translated-->
-    <value>&amp;Emplacement actuel cible</value>
+    <value>&amp;Cibler l'emplacement actuel</value>
     <source>&amp;Target current location</source>
   </data>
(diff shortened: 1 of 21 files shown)


SrvSurvey/forms/FormGroundTarget.fr.resx
```diff
@@ -485,12 +485,12 @@
   </data>
   <data name="btnTargetCurrent.Text">
     <!--Machine translated-->
```
njthomson, line 487: In many files you have added a manual translation but left the `<!--Machine translated-->` comment - could you remove these?

njthomson (commented):
@Entrivax - apart from needing to remove `<!--Machine translated-->` where you have added manual translations - these changes look good.

Entrivax:
It should be good now, some of the one I had left with the comment were just one word fix without a proper check of the rest of the sentence. And I forgot to remove it on the rest.

njthomson (approved):
Nice work!

PR #613, opened 2025-11-28
Added color scheme for first discoveries and already analyzed bios

jmyoung:
Added two additional colors to VolColors, and control logic for the BioSystem plotter to differentiate between some extra scenarios;

- Bio is a first discovery and has been analyzed:  Show bio colored in gray with golden border
- Bio has been analyzed and is not a first discovery:  Show bio colored in gray

This then provides additional feedback to the CMDR about the status of signals in a system, making it clearer which signals haven't been collected, which have, and which were first discoveries for that commander, without losing the ability to see those first discoveries after the fact while flying around.  Example;

image

In that screenshot, 6d signal 1 is an uncollected signal that is not a first discovery.  Signal 2 is a collected signal that is not a first discovery.  Signal 3 is a collected signal that was a first discovery.  The fourth signal on 8f is a first discovery that has not been collected yet.


jmyoung:
Updated to keep track with main.

njthomson:
@jmyoung - I love and appreciate that you have made this PR - encapsulating all the states in such a small amount of screen space is a challenge and clever use of colours is exactly a way to be successful here. However it presents a mild challenge. I have found that many users do not notice the more subtle differences in colours or they find it frustrating having to infer the significance. I tried a broader range of colours whilst working on early Colonisation features and the feedback was that it was too confusing to follow.

That said, I do not wish to lose the extra specificity you have enabled but I am hesitant to turn it on for everyone without trying it out. I'm also keen to adapt the colours used into the theming plumbing that I am slowly (glacially slowly) migrating overlays to use, as can be seen at the bottom of [this commit](https://github.com/njthomson/SrvSurvey/commit/3f01f1743a06c670fc06f6272fa6e4fb9ceb3083#diff-4e612a5f2c7a9c974f873cbeb81f334d5f645e1b95e0219683bacff5d626b466) ... but that is beyond the scope of what you intended for this PR.

I think what I would like to do is directly incorporate your changes as I adapt VolCol to use theming from C.cs, only I'll make it that the new colours only kick in when a debugger is attached (or I can add a hidden setting) - that way you can continue to play with the revised colours and I can try it out before unleashing it on everyone.

jmyoung:
@njthomson Yes I had thought about that. A hidden setting until it’s cleaner could be good. I also thought about whether there is a clean way to take the existing themed Orange colour and darken it a few stops and do the same with the existing gold, so that the colour scheme is preserved when customised. 

Eg in that case the logic would just boil down to bright shades for uncollected states, and dark shades for collected states. 

Thoughts?

jmyoung:
@njthomson I refactored the changes to instead produce dim versions of the existing (and customizable) color schemes, and also included a checkbox in settings that defaults to off that can be used to enable/disable the dimming.

Example results;

image

6d signal 1 = unanalyzed signal, signal 2 = analyzed signal, signal 3 = analyzed first discovery signal.  8f signal 4 is an unanalyzed first discovery signal.

The widget in the settings window looks like this;

image

Hopefully that result will be more intuitive, and sticks more with the color scheme as it is, just taking advantage of dimming to reduce the emphasis on specific signals if they are analyzed.

njthomson:
@jmyoung - your changes look okay, it's just hard to see what is changing on the settings form.
(Things are rather cramped in there in general and it's easy for things to break in high-dpi)

jmyoung:
@njthomson Apologies, I was just fixing the form, I wasn't expecting it to message you straight away (I'm a bit weak with the small details of how Git works.  I was fixing up the merge conflict (caused by your new checkbox using checkBox53, so I undid what I'd done to bring back in your changes and then re-did the work with a new name so it wouldn't conflict.  Should be fine now.

njthomson:
No worries at all :)

njthomson:
The diff's in FormSettings are still broader than I would have expected (Winform's fault, not yours). How about:
A) Revert your changes in that file
B) Let me add the checkbox right afterwards:
image

jmyoung:
> The diff's in FormSettings are still broader than I would have expected (Winform's fault, not yours). How about: A) Revert your changes in that file B) Let me add the checkbox right afterwards: image

Actually that's a fantastic suggestion.  I've just been fighting with it trying to get it clean, and it's not really playing ball.  Probably best if you add it :D  Give me a minute to revert those changes and push.

jmyoung:
@njthomson Done.  The overall pull diff now looks a lot cleaner indeed.  Winforms certainly can make a mess, lesson learned.

jmyoung:
No worries at all :)

njthomson:
@jmyoung - your changes are brilliant. I did a little exobiology with that new setting applied and it feels completely natural and obvious. I am inspired to make the setting default to `on` for everyone with the next check-in.

Thank you again for your efforts ❤️

PR #698, opened 2026-01-21
Fix typo Interstellar Factions -> Interstellar Factor

orangejulius:
There's what I assume is a repeated typo of the contact where you can turn in bounties, pay your own fines/bounties, etc. This should fix it

https://elite-dangerous.fandom.com/wiki/Interstellar_Factor

### DIFF
--- a/SrvSurvey/plotters/PlotHumanSite.Designer.cs
+++ b/SrvSurvey/plotters/PlotHumanSite.Designer.cs
@@ -187,7 +187,7 @@ namespace Loc {
         }
         
         /// <summary>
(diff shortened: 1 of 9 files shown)


njthomson:
Good catch @orangejulius! 

As a string resource is changing, my home-grown loc process should be rerun (so that machine translations for other languages get updated) but ... this process can only be run from dev-box (as it needs access to private Azure resources)

I am grateful for your PR, but let me repeat the change, run that process and push the changes.

njthomson:
Thank you again @orangejulius - I've since made the changes and this PR is no longer needed.

orangejulius:
Awesome, glad to help :)

PR #791, opened 2026-01-22
fix: prevent bodyScans double-counting from DSS scans

### LINKED ISSUE #204: SystemData::Counts::bodyScans can have a higher number than bodyCount
`bodyScans` is incremented every time a `Scan` event occurs
https://github.com/njthomson/SrvSurvey/blob/3711760341bc6f8c5b73778dd4c33b7417290c87/SrvSurvey/game/CommanderJourney.cs#L270
`bodyCount` is set from the `FSSDiscoveryScan` event

All seems to be well, however, there is no check that the body has been scanned before already. Most common example is a DSS scan - it creates an extra `Scan` event (alongside it's own `SAAScanComplete`), even though there was one after a `Detailed` or `AutoScan` `Scan`

This is the result

Image

teal-bauer:
DSS creates a second Scan journal event for bodies already scanned with FSS. The `onJournalEntry(Scan)` handler increments `bodyScans` unconditionally, so DSS-ing a body counts it twice.

Track scanned BodyIDs in a HashSet, only increment on first encounter.

Fixes #204

### DIFF
--- a/SrvSurvey/game/CommanderJourney.cs
+++ b/SrvSurvey/game/CommanderJourney.cs
@@ -278,7 +278,11 @@ namespace SrvSurvey.game
             // ignore asteroid clusters
             if (entry.PlanetClass == null && entry.StarType == null) return;
 
-            // Effectively (but not entirely) FSS scans
+            // track which bodies we've counted to avoid double-counting from DSS scans
+            currentSystem.bodiesScanned ??= new();
+            if (currentSystem.bodiesScanned.Contains(entry.BodyID)) return;
+            currentSystem.bodiesScanned.Add(entry.BodyID);
+
             currentSystem.count.bodyScans++;
 
             if (entry.StarType != null)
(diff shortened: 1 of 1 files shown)


teal-bauer:
I _thought_ FSS scan events couldn't occur when you've scanned a body once, but I also had the game forget star system data I discovered a few times - however I don't remember if it was on the same machine and in the same "lifetime", so I'm just not sure. It would certainly make sense to only ever get one DSS and one FSS scan event from the game but… well :) I will look through my historical journals when I'm back at my desk, see if there's any duplicate cases over multiple sessions

teal-bauer:
For the question of what data to persist and for how long, I don't see any good workaround. Some other tools just parse all journal files every time they start and keep state only in memory - this has some benefits but also a bunch of drawbacks. I was also thinking that an expiry system might be good, but it would require rearchitecting the persistence layer to use e.g. SQLite and something like event sourcing, or at least storing timestamps on records. Unless the "naive" way of just deleting CommanderJourneys older than a certain timestamp is acceptable…?

teal-bauer:
Checked 121 journal files (248,981 events). Found 543 bodies with duplicate Scan events:

- **291**: Approach scan followed by DSS probe scan (normal)
- **111**: Left body in supercruise, came back, rescanned
- **72**: Revisited system after game restart
- **68**: Nav beacon data followed by approach scan
- **24**: Player died, respawned, rescanned same bodies
- **3**: Flew to different body, returned, rescanned

So yes, the game does send duplicate Scan events across sessions. HashSet deduplication handles all these cases.

PR #837, opened 2026-01-26
feat: show neutron indicator in jump overlay for Spansh routes

### LINKED ISSUE #673: Feature request - add REFUEL indicator in jump route overlay
When importing a route from spansh plotter link in the clipboard, the relevant window reports all systems where one needs to refuel. But since that window is not an overlay, when the focus moves to the game it disappears. Consider adding the REFUEL indicator to the jump overlay (the one in normal gameplay that appears when initiating a jump) so that players with only one screen available don't need to juggle between windows and to meddle with window visibility settings.

teal-bauer:
Follow-up to #834. Shows the neutron indicator in the jump overlay for Spansh routes — the star you need to supercharge at isn't always the primary, so having the label there helps.

Now also covers PlotSphericalSearch, using the same style as notes display. Both refuel and neutron labels use FormRouteExtras localized strings, matching FormRoute.cs.

See also #180, which uses a computed alerts property on Hop.

Fixes #673

### DIFF
--- a/SrvSurvey/plotters/PlotSphericalSearch.cs
+++ b/SrvSurvey/plotters/PlotSphericalSearch.cs
@@ -1,4 +1,5 @@
 ﻿using SrvSurvey.game;
+using SrvSurvey.Properties;
 using SrvSurvey.widgets;
 using Res = Loc.PlotSphericalSearch;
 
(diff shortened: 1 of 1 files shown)


teal-bauer:
I realized #180 also does this, and that approach seems a bit cleaner with the separate `.alerts` property and localization. Would you prefer to merge that instead? Otherwise I can extend this with localization and spherical plot support. Sorry for the duplication.

njthomson:
> I realized #180 also does this, and that approach seems a bit cleaner with the separate `.alerts` property and localization. Would you prefer to merge that instead? Otherwise I can extend this with localization and spherical plot support. Sorry for the duplication.

Go for it - that would be a nice addition.

teal-bauer:
Added PlotSphericalSearch support — same style as the notes display, with localized labels.

PR #850, opened 2026-01-30
feat: move satellite assemblies into locales/ subfolder

teal-bauer:
## Summary

- Filter NuGet-only locale folders (cs, it, ja, ko, pl, tr, zh-Hant) via `SatelliteResourceLanguages` — they contained no app resources
- Post-build inline task moves remaining culture folders into `locales/`, decluttering the output directory
- `AssemblyResolve` handler in `Program.Main()` resolves satellite assemblies from `locales/<culture>/`
- Replace `cmd.exe` and `PowerShell` build steps with cross-platform MSBuild equivalents (`<Copy>` task, `RoslynCodeTaskFactory` inline C#)

### Technical Notes

The `MoveSatelliteAssemblies` task dynamically finds directories containing `*.resources.dll` — no hardcoded language list. `SatelliteResourceLanguages` in the `<PropertyGroup>` is the single source of truth for supported locales.

The `SortResxFiles` task is a direct port of `sort-resx.ps1` to inline C#, so it runs on any platform without requiring PowerShell.

### DIFF
--- a/SrvSurvey/Program.cs
+++ b/SrvSurvey/Program.cs
@@ -47,9 +47,17 @@ namespace SrvSurvey
         [STAThread]
         static void Main(string[] args)
         {
-            // To customize application configuration such as set high DPI settings or default font,
-            // see https://aka.ms/applicationconfiguration.
-            //ApplicationConfiguration.Initialize();
+            AppDomain.CurrentDomain.AssemblyResolve += (sender, args2) =>
+            {
+                var name = new AssemblyName(args2.Name);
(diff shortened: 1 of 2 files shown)


teal-bauer:
Feel free to apply only parts - since I am often building this on Linux the PowerShell stuff was a bit of a nuisance.

njthomson:
The need to alpha sort .resx files is only relevant when resources are changing. I am inclined to make another build configuration that excludes it, so general dev builds become faster.

teal-bauer:
Yeah I wonder if we can make that dependent on whether the resources changed somehow, but I'm not too familiar with MSBuild, I had to read docs for this already...

PR #881, opened 2026-02-09
Anchor FSS progress percentage to in-game honk progress

teal-bauer:
## Summary

- Use the game's `FSSDiscoveryScan.Progress` value as baseline for FSS completion percentage
- Interpolate linearly toward 100% as bodies are scanned after honking
- Store `honkProgress` and `honkBodyCount` in SystemData for the calculation

Previously, FSS progress was `fssBodyCount / bodyCount` — a flat ratio that significantly underreported compared to the in-game progress bar. In a 12-body triple-star system, SrvSurvey showed 16% after honking while the game reported ~50%.

The game's progress is signal-based (some bodies represent larger chunks of the FSS spectrum), and the `Progress` field from `FSSDiscoveryScan` is the only time we get the game's own value. This change anchors to that, then distributes the remaining progress evenly across bodies left to scan.

### Before/after (Eishaw SK-K b42-0, 12 bodies, 3 stars)

| Moment | Before | After | In-game |
|--------|--------|-------|---------|
| After honk (2 stars auto-scanned) | 16% | 50% | ~50% |
| After FSS'ing 5 more bodies | 58% | 75% | ~75% |
| All bodies found | 100% | 100% | 100% |

### Files changed

- `SystemData.cs`: Add `honkProgress`/`honkBodyCount` fields, `fssProgress` computed property
- `PlotSysStatus.cs`: Use `fssProgress` instead of inline body-count ratio

## Test plan

- [x] Build succeeds (`dotnet build SrvSurvey.sln`)
- [ ] Honk a multi-star system, verify overlay % is close to in-game FSS progress
- [ ] Complete FSS, verify 100% / checkmark appears as before
- [ ] Visit a previously-honked system, verify progress displays correctly from persisted data

### DIFF
--- a/SrvSurvey/game/SystemData.cs
+++ b/SrvSurvey/game/SystemData.cs
@@ -553,6 +553,14 @@ namespace SrvSurvey.game
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore, DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
         public bool fssAllBodies;
 
+        /// <summary> The game's FSS progress (0.0-1.0) at honk time </summary>
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore, DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        public double honkProgress;
+
+        /// <summary> How many bodies were already known (auto-scanned) at honk time </summary>
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore, DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        public int honkBodyCount;
+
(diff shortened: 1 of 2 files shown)


teal-bauer:
This is definitely not perfect but it should be quite a bit closer to the game's number. The reason is that a system's stars account for the majority of the mass. That means if we take the progress number from the FSS honk as our reference, our error will become smaller.

njthomson:
Alas ... I originally intended the percentage to be strictly by bodies scanned. In high body-count systems with many gas giants, I found it annoying that the game would get to ~90% with lots of moons still to go. I have no qualms with this code being submitted, but there should be a setting telling the UI which calculation method to use.

I can help adding that setting to the form (if the form designer/code generator goes berserk that is, it is super sensitive on that one) ... or we could leave this as a hidden setting.

Issue #875, opened 2026-02-06
Suggestion: planet visit order optimizer


Jessidhia:
One of the things the Elite Observatory Evaluator plugin can do is display an optimized order to visit mapping/landing candidate planets:

<img width="447" height="175" alt="Image" src="https://github.com/user-attachments/assets/dad2bc33-a385-450c-b7e7-fdf95a8d47ee" />

These plugins seem to be closed source so there's not an easy way to reference the algorithm, but Elite only uses simple Keplerian mechanics for orbits so [the orbit can be entirely described](https://en.wikipedia.org/wiki/Orbital_elements#Classical_Keplerian_elements) by the `Parents`, `SemiMajorAxis`, `Eccentricity`, `OrbitalInclination`, the angle of the `AscendingNode` and the longitude of `Periapsis`; the position within the orbit can be determined from the `MeanAnomaly` (though it isn't defined for eccentricity 1 orbits, so it's possible this is a misnomer for mean longitude)

<img width="1157" height="969" alt="Image" src="https://github.com/user-attachments/assets/bf3100a3-0f50-41c6-8ec5-ea1b77147365" />

I... do not have enough math in my head to actually find an algorithm for this, though. The high level idea is to use orbital parameters + `MeanAnomaly` to find the position of the body at FSS time (adding 360°/`Period` \* seconds since FSS to keep the `MeanAnomaly` current) to convert them to 3D positions, then plot the shortest total distance between the ship's current position (probably unavailable, use center of A star instead?) visiting all candidates.

One thing that can be done with Evaluator, though, is adding or removing planets to visit from the list, whether manual or through input from additional plugins, while SrvSurvey's list is entirely determined by mapping value.

teal-bauer:
We need to store the data so we can use it later. The actual Keplerian mechanics should be straightforward... I'll have a go at this.

teal-bauer:
Turns out the two-body problem is much simpler than the three-body problem ;) I should have a PR for this tomorrow I think o7

teal-bauer:
Ok I said "straightforward" but of course this includes a) a system of equations that cannot be analytically solved and b) a traveling salesman problem ... heuristics to the rescue! ;)

teal-bauer:
It's taking longer than expected to find test systems for this as I'm out at Beagle Point with DW3 :D

teal-bauer:
[...]

PR #884, opened 2026-02-10
feat: planet visit order optimizer

teal-bauer:
Adds orbital position calculation and route optimization for DSS targets, as described in #875.

## Changes

- **OrbitalMechanics.cs** (new): Keplerian orbit solver (Newton-Raphson for eccentric anomaly) + perifocal-to-inertial 3D positioning with recursive parent accumulation. Double-precision vectors because `System.Numerics.Vector3` loses precision at 100+ AU distances. Route optimizer uses exact permutation search for ≤10 targets, nearest-neighbor + 2-opt for larger systems.
- **SystemData.cs**: `getDssRemainingBodyIds()` reuses the existing DSS skip filters. `CalculateOptimalRoute(targetBodyIds)` builds the orbital model and sets `visitOrder` on each body.
- **PlotSysStatus.cs**: Shows visit route as `1 > 3 > 5 > 5a > 4 > 2 > 6` with destination body highlighting. DEBUG builds get a diagnostic overlay with hop distances and route comparison.

Implements #875

## Status

This is a first working version — tested manually in a handful of systems, route looks correct compared to the orrery view. Still needs more testing with edge cases (missing orbital data, barycentre-heavy systems, very large systems with >10 DSS targets).

### DIFF
--- a/SrvSurvey/game/SystemData.cs
+++ b/SrvSurvey/game/SystemData.cs
@@ -675,6 +675,13 @@ namespace SrvSurvey.game
             body.mass = entry.MassEM > 0 ? entry.MassEM : entry.StellarMass; // mass
             body.distanceFromArrivalLS = entry.DistanceFromArrivalLS;
             body.semiMajorAxis = entry.SemiMajorAxis;
+            body.eccentricity = entry.Eccentricity;
+            body.orbitalInclination = entry.OrbitalInclination;
+            body.periapsis = entry.Periapsis;
+            body.meanAnomaly = entry.MeanAnomaly;
+            body.orbitalPeriod = entry.OrbitalPeriod;
+            body.ascendingNode = entry.AscendingNode;
+            body.orbitalEpoch = entry.timestamp;
             body.absoluteMagnitude = entry.AbsoluteMagnitude;
             body.radius = entry.Radius;
             body.parents = entry.Parents;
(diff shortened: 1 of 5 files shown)


Jessidhia:
Out of curiosity, I wonder if they gave the Voyager probe POIs in the Sol system hyperbolic orbits (negative eccentricity). Probably not something the optimizer would actually encounter, though, since Elite otherwise doesn't do comets 😞

It is also unclear what happens to the argument of Periapsis / MeanAnomaly when something has an eccentricity of *exactly* 1, or AscendingNode when the OrbitalInclination is exactly 0, but I don't know what body / system can be used to check for it.

teal-bauer:
Didn't Elite use to do comets? But yeah, I doubt they'd special case for something small like that. Would be interesting to log this data and see if any systems have such borderline values.

I've been doing some more testing and I think this only needs a UX decision before it's ready to merge - right now I show the DSS list and then the visit order underneath - @njthomson Do you think we should just merge that into one, or keep it separate?

teal-bauer:
Example what it looks like rn:
image
image

njthomson:
I am back from my travels ... let me look at the code and ponder the UX design.

SrvSurvey/game/SystemData.cs
```diff
@@ -1910,6 +1940,94 @@ public async Task<double> getNebulaDist()
...
+                    if (body != null)
+                        body.visitOrder = i;
+                }
+
+                lastCalculator = calculator;
+                Game.log($"Calculated optimal route for {targetBodyIds.Count} bodies in {this.name}");
+            }
+            catch (Exception ex)
+            {
+                Game.log($"Error calculating optimal route: {ex.Message}");
```
njthomson: May I recommend including StackTrace, so it's automatically written to log files:
```
$"Error calculating optimal route: {ex.Message}\r\n\t{ex.StackTrace}"
```

SrvSurvey/game/SystemData.cs
njthomson: Ideally (or in a subsequent PR) we would populate orbital mechanic numbers from EDSM.
(Assuming it has all of them?)

njthomson:
@teal-bauer - if you can get the main entry points shielded behind a new setting, I see no reason not to get this in and try it out.

teal-bauer:
Aye! I hope to find some time on the weekend!

Fenris159:
A replacement draft is now open at #1023, based on current main. It preserves @teal-bauer original two commits and authorship, and adds a separate hardening commit for the owner feedback and review findings discussed here. Keeping #884 as the original design and review history; maintainers can decide when to close it as superseded.

PR #933, opened 2026-03-28
Fixes to bio prediction

davidtang123:
Revert to constant modifications, corrections for aleoida laminiae + spica, bacterium tela rocky ice argon, and all bio that use "orion-cygnus arm".


njthomson (approved):
Nice work!

PR #1019, opened 2026-07-19
Improvement of Sinuous Tubers predictions, fix to Bacterium Tela Helium atmos, fix for regional first discovery highlighting

davidtang123:
Tubers will now only predict if within a known Tubers bubble or Empyrean Straits to reduce overprediction.

Bacterium Tela - Helium will now correctly predict when under the known conditions.


Fenris159 (changes requested):
The Bio Criteria changes look great however the regional-first implementation still has two edge cases that need addressing.

### 1. Preserve `isNewEntry`

`CodexEntry` updates the regional commander Codex before setting `organism.isNewEntry`. The new `isFirst` getter then sees the entry as already discovered in the region and no longer considers `isNewEntry`.

For a variant previously discovered in another region, this causes the qualifying regional-first highlight to disappear immediately after discovery. Please preserve `isNewEntry` when `highlightRegionalFirsts` is enabled.

Relevant code:

- [Regional Codex update](https://github.com/njthomson/SrvSurvey/blob/dd266d9d17762c2608e4deb4841bf108f439d70a/SrvSurvey/game/SystemData.cs#L1014-L1021)
- [Journal flag update](https://github.com/njthomson/SrvSurvey/blob/dd266d9d17762c2608e4deb4841bf108f439d70a/SrvSurvey/game/SystemData.cs#L1083-L1093)
- [New `isFirst` getter](https://github.com/njthomson/SrvSurvey/blob/dd266d9d17762c2608e4deb4841bf108f439d70a/SrvSurvey/game/SystemData.cs#L2760-L2779)

### 2. Guard unresolved entry IDs

`isPersonalFirstDiscovery` returns `true` when no matching entry exists. Because the new getter can call it with `entryId == 0`, unresolved genus-only organisms can be classified as commander firsts.

The collapsed Predictions overview evaluates `isFirst` for every organism, so this can incorrectly highlight an ordinary body. Please validate the entry ID before performing the discovery checks.

Relevant code:

- [`isPersonalFirstDiscovery` behavior](https://github.com/njthomson/SrvSurvey/blob/dd266d9d17762c2608e4deb4841bf108f439d70a/SrvSurvey/game/CommanderCodex.cs#L137-L148)
- [Collapsed Predictions overview](https://github.com/njthomson/SrvSurvey/blob/dd266d9d17762c2608e4deb4841bf108f439d70a/SrvSurvey/forms/FormPredictions.cs#L638-L649)

The persistent regional-Codex lookup is the right addition, but it needs to be combined with the existing journal flag and an entry-ID validity check.

njthomson:
Hi @davidtang123 - I had a chance to look over your changes finally. These look good and I've no further comments beyond what Fenris already asked - but these do need to be addressed still.

davidtang123:
I've added changes that should address the above. Holding off on other planned changes (atmospheric Horizons bio) that are not included with this update.

PR #1023, opened 2026-07-28
feat: planet visit order optimizer

Fenris159:
## Summary

This draft supersedes #884 by carrying the planet visit-order work onto the current `main`. It preserves @teal-bauer's original two commits and authorship, then adds a separate hardening commit based on the review discussion in #884.

- adds a **default-off** `Show optimized DSS route` setting around the route UI and calculation entry point
- uses one DSS target-selection path for both the existing display and the optimizer, including ring destinations
- reconstructs nested barycentre parents from journal parent chains and supports a root barycentre in multi-star systems
- rejects missing, cyclic, or invalid orbital data instead of treating incomplete bodies as if they were at the origin
- replaces factorial permutation search with exact Held-Karp optimization through 15 targets, then nearest-neighbor plus 2-opt for larger systems
- caches rendered routes briefly and invalidates them when new orbital scans arrive
- includes stack traces in route-calculation error logging
- adds focused regression and performance-boundary coverage to CI

## Review history

- Supersedes #884; that PR remains the source of the original implementation and review discussion.
- Continues the feature requested in #875.
- Spansh orbital enrichment is staged separately in [Fenris159/SrvSurvey#2](https://github.com/Fenris159/SrvSurvey/pull/2), based on this branch so its diff remains isolated.
- EDSM can provide supplemental static orbital metadata, but it cannot independently reconstruct current positions because its bodies API does not expose ascending node or mean anomaly.
- Manual route target editing and external-plugin target input remain follow-ups rather than blockers for this draft.

## Validation

- `dotnet test tests/SrvSurvey.OrbitalTests/SrvSurvey.OrbitalTests.csproj --configuration Release --no-restore` — 8 passed
- `dotnet build SrvSurvey/SrvSurvey.csproj --configuration Release --no-restore` — succeeded with 0 errors (existing repository warnings remain)

Runtime verification in Elite Dangerous is still needed for the final overlay layout and real journal-system coverage.

### DIFF
--- a/.github/workflows/dotnet-desktop.yml
+++ b/.github/workflows/dotnet-desktop.yml
@@ -75,7 +75,12 @@ jobs:
     - name: Install .NET Core
       uses: actions/setup-dotnet@v4
       with:
-        dotnet-version: 7.0.x
+        dotnet-version: |
+          7.0.x
+          8.0.x
+
+    - name: Test orbital route calculations
+      run: dotnet test tests/SrvSurvey.OrbitalTests/SrvSurvey.OrbitalTests.csproj --configuration Release
 
     # Add  MSBuild to the PATH: https://github.com/microsoft/setup-msbuild
     - name: Setup MSBuild.exe
--- a/SrvSurvey/Settings.cs
+++ b/SrvSurvey/Settings.cs
@@ -110,6 +110,7 @@ namespace SrvSurvey
         public int hideFssLowValueAmount = 10_000;
         public bool skipHighDistanceDSS = false;
         public int skipHighDistanceDSSValue = 100_000;
+        public bool showDssVisitRoute = false;
         public bool showNonBodySignals = false;
         public bool autoTrackCompBioScans = true;
         public bool skipAnalyzedCompBioScans = true;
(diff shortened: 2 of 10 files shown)


Fenris159:
The Spansh-only orbital enrichment follow-up is now staged at https://github.com/Fenris159/SrvSurvey/pull/2. It is based on this PR so the five-file enrichment diff stays isolated. It consumes complete timestamped Spansh orbit snapshots, fixes AU and day unit conversion, and keeps local journal scans authoritative. EDSM is intentionally treated only as supplemental static metadata.

Fenris159:
After comparing this replacement with the original goals in #875, the implementation and discussion in #884, and @njthomson’s review feedback, my assessment is:

This completes the automatic DSS visit-order optimizer scope. The feature is default-off, shares the existing DSS filtering logic, handles barycentres and invalid orbital data safely, uses the commander’s current body when available, and includes the requested error logging and test coverage.

The stacked Spansh follow-up at Fenris159/SrvSurvey#2 also uses the correct existing data-consumption path. It accepts only complete, timestamped orbital snapshots and keeps journal data authoritative. EDSM cannot provide equivalent enrichment because it lacks ascending node and mean anomaly.

It does not yet cover every capability discussed in #875:

- manually adding or removing route targets;
- including landing-only or biological destinations after DSS is complete;
- external-plugin target input;
- combining the existing DSS list and optimized order into a single selectable list.

I do not consider those blockers for this default-off first phase. My recommendation is to merge the automatic DSS optimizer first, follow with the Spansh enrichment after rebasing it onto upstream `main`, and track manual/landing target selection separately.

The remaining requirement before calling the orbital implementation fully proven is live testing against several Elite systems and their orrery views.

PR #1025, opened 2026-07-28
Add opt-in Inara commander uploads

Fenris159:
## Summary

- make Inara a regular nullable member of one initialized `Game`, with no process-wide singleton or active/last-commander identity lookup inside the integration
- bind each upload session to the exact `CommanderSettings` name and FID, derive eligibility once from the validated game version, and read only that profile's API key; keys and queued events never carry across commander sessions
- use API-key presence as the per-commander opt-in, keep journal-derived mapper state warm before opt-in, and discard old queued data when a key is replaced or cleared
- batch requests every 35 seconds with a 20-second request timeout, one Inara-owned bounded retry policy (with no automatic POST retries underneath it), strict response-shape validation, and credential-safe diagnostics
- coordinate journal ingestion, active sends, final credit reporting, one bounded final flush, and transport disposal during every `Game` session teardown
- preserve Live/non-beta/non-multicrew eligibility and use `Elite.hadManyGameProcs` to suppress ambiguous shared sidecar/status enrichment without new process enumeration
- pin the settings form to one captured commander profile, display the commander as read-only, mask the key, and provide `OK` and confirmed `Clear Key` actions
- hard-code `isBeingDeveloped = true` for the current Inara application-validation phase; there is no persisted developer-mode setting
- update privacy documentation, add commit-pinned EDMarketConnector attribution, include the rewrite specification, and expand focused session/transport regressions

## Session and credential design

One `Game` owns zero or one `Inara` instance. Creation occurs only after journal initialization has established the first `Fileheader`, commander settings, commander name, FID, and game version. The instance captures that identity and reads the API key only from the same bound `CommanderSettings` object.

Every queued event carries the API-key snapshot under which it was accepted; the session-owned queue and mapper already provide the fixed commander/FID boundary. Replacing or clearing the key discards events accepted under the old key. Changing commanders disposes and finalizes the old `Game` and its Inara instance before the new session is constructed, so mapper state, credit state, queues, timers, tasks, and credentials cannot cross that boundary.

The Inara commander name is the regular Elite commander name stored in `CommanderSettings.commander`; it is not independently editable. Inara associates the request through the commander name, Frontier ID, and that commander's personal API key.

Finalization runs on every `Game` session disposal, including commander changes?not only on application shutdown. Clearing a key is a distinct opt-out path: pending uploads are discarded and no final request is made with the cleared key.

## Behavior retained

- supported journal data is translated into Inara events rather than forwarding raw journal entries
- credits start from exact `LoadGame` values, use journal transaction deltas, refresh assets from `Statistics`, and report at session boundaries or no more often than hourly
- the current journal is replayed locally with uploads disabled before live callbacks begin, so history seeds state without being uploaded
- current ship cargo may be seeded from `Cargo.json`; shared `Cargo.json`, `ShipLocker.json`, and `Status.json` data is suppressed after multiple Elite processes have been observed
- Legacy, beta/alpha, and multicrew uploads remain suppressed
- no generic application API token is embedded; each commander supplies a personal Inara API key

## Validation

- `dotnet test SrvSurvey.Tests/SrvSurvey.Tests.csproj --arch x86 -p:PlatformTarget=x86 --no-restore` ? passed, 62 tests
- focused cross-session request capture verified commander name, FID, API key, and event data always come from one bound session
- `dotnet build SrvSurvey/SrvSurvey.csproj -c Release -p:PlatformTarget=x64 --no-restore` ? passed with 0 errors; existing repository warnings remain
- `git diff --check` ? passed
- the full solution also compiled the app and tests in Release x64, then stopped at `SrvSurvey.Package` because this environment does not contain `Microsoft.DesktopBridge.props`
- the settings flow is covered by captured-profile regressions and designer inspection; no live Inara write was executed because that requires a personal test commander and mutates its Inara profile
- no live two-client multibox smoke test was executed; focused inventory, sidecar, taxi-attribution, and session-isolation regressions cover the implemented safeguards

[Inara API development guide](https://inara.cz/elite/inara-api-devguide/)

[Inara API documentation](https://inara.cz/elite/inara-api-docs/)

[Commit-pinned EDMarketConnector Inara implementation](https://github.com/EDCD/EDMarketConnector/blob/2b6a0ce1ee3ba60c21f3f4e9fa093046da8825e4/plugins/inara.py)

[Rewrite specification](https://github.com/Fenris159/SrvSurvey/blob/agent/inara-upload/docs/specs/inara-session-integration-rewrite.md)


njthomson:
- I think it would make sense to make `inaraDeveloperTestMode` a hidden setting. I imagine virtually zero regular users would need to set it, and those few people who do can hand edit `%appdata%\SrvSurvey\SrvSurvey\1.1.0.0\settings.json` easily enough.

- The *External data* tab in settings is chocka full. I'm going to move the Raven Colonial api-key entry to another location, so you can use that space (and that group-box) for the Inara. This then means you won't have to shift that Canonn mini screenshot around.

I plan to move the Raven Colonial api-key very soon.

Fenris159:
That would be fine. It can just as well be left on by default because Artie is enforcing it anyway unless further whitelisted for the extra data gated by that flag. `inaraDeveloperTestMode ` only gates data collected related to CG and CG goal completion. Therefore it's of little consequence.

njthomson:
@Fenris159 - if you pull latest, I've submitted my changes on the settings form. Rather than having the api-key edited inline, I recommend adding a new little window for it, like I just did for the Raven api-key.

SrvSurvey/Settings.cs
```diff
@@ -281,6 +281,14 @@ class Settings
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore, DefaultValueHandling = DefaultValueHandling.Ignore)]
         public string? eddnEnvironment = null;
 
+        /// <summary>Whether the user has explicitly opted in to Inara commander updates.</summary>
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore, DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public bool inaraUpload = false;
+
+        /// <summary>Whether Inara should treat uploads as development or test data.</summary>
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public bool inaraDeveloperTestMode = true;
```
njthomson: The effect of this line is as I shared on Discord. When setting `inaraDeveloperTestMode` is not known, SrvSurvey will initialize it with the default value here. If we change it to `false` later on, it will have no impact for anyone who already has it - only for new users who never had the setting would see a difference.

I think this should default to `false` and let people who know what they are doing change it to `true` when they know it is necessary. *(That Artie has it hard-coded on his end also makes this a safe option)*

Or ... if you think it really is good to start that way, but we would change it later, make it `inaraDeveloperTestModeTEST = true;` and we can rename it to `inaraDeveloperTestMode = false;` later on.

SrvSurvey/net/InaraEventMapper.cs
```diff
@@ -0,0 +1,897 @@
+using Newtonsoft.Json.Linq;
+
+// Behavioral reference: EDMarketConnector's plugins/inara.py and monitor.py.
```
njthomson: Could you quote a permalink to this?
*(and same comment for the other files based on EDMC code)*

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+    /// </summary>
+    internal sealed class Inara : IDisposable
+    {
+        internal const string Endpoint = "https://inara.cz/inapi/v1/";
+        private static readonly TimeSpan sendInterval = TimeSpan.FromSeconds(35);
+        private readonly HttpClient? client;
+        private readonly InaraEventMapper mapper = new();
+        private readonly InaraEventQueue queue = new();
+        private readonly System.Threading.Timer? timer;
+        private Game? currentGame;
```
njthomson: This will cause problems. The instance of `Inara` is static and long lived, whereas a `Game` instance is destroyed and recreated anytime someone returns to the main menu.

You could make it a getter to the static singleton:
`private Game? game => Game.activeGame;`

This will also prevent `Game` instances from getting GC'd as this reference would live for ever.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+
+        public Inara()
+        {
+            HttpClient? configuredClient = null;
+            System.Threading.Timer? configuredTimer = null;
+            try
+            {
+                configuredClient = new HttpClient(Util.getResilienceHandler())
+                {
+                    Timeout = TimeSpan.FromSeconds(20),
```
njthomson, line 82: Should this duration be a hidden setting, defaulting to 20s? *(This way it can be changed without needing a new build)*

Unless Artie says to make it always be 20s ... I would make it a hidden setting.

Same comment for `sendInterval` above. I'm also curious why we want the values to be different?

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+                        : null;
+                }
+
+                var seededCount = SeedState(
+                    mapper,
+                    entries,
+                    createContext(game, canPrepareUpload && !multiboxing),
+                    cargoInventory);
+                Game.log($"Inara seeded current state from {seededCount} journal event(s).");
+                if (multiboxing)
```
njthomson, line 128: Food for thought: *(for another PR that is)*

If the timestamp in *cargo.json* matches the timestamp from the `Cargo` event ... that indicates it is safe to read the file, that it has not been clobbered by another process. Alas, this won't work for *status.json* and I've not worked with *shiplocker.json* enough to have an opinion there.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+                    Game.log("Inara multi-box mode: shared Cargo.json, ShipLocker.json, and Status.json data is suppressed.");
+            }
+            catch (Exception ex) when (ex is IOException or UnauthorizedAccessException or JsonException)
+            {
+                Game.log($"Inara could not seed current journal state ({ex.GetType().Name}).");
```
njthomson: I would log the exception message and call-stack, unless you really don't care to know why it is failing?

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+
+            RunIsolated(
+                () => onJournalEntryCore(game, raw),
+                ex => Game.log($"Inara ignored {raw["event"]?.ToString() ?? "unknown"} without affecting other journal processing ({ex.GetType().Name})."));
+        }
+
+        private void onJournalEntryCore(Game game, JObject raw)
+        {
+            // Manual calls made while Game reconstructs state from journal history must never upload.
+            if (!Game.ready || Game.activeGame != game) return;
```
njthomson: > Game.activeGame != game
This feels problematic. I would go with `Game.activeGame == null` 

... or `this.game == null` if you honoured my suggestion above to make this.game be a getter

njthomson: I would also include `|| !Game.settings.inaraUpload` ... so we exit early as possible if the setting is off.

Though, would that cause problems if a user then toggles it to on part way through a session?

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+                    currentGame = game;
+                    Game.log($"Inara startup seeding was skipped without affecting SrvSurvey ({ex.GetType().Name}).");
+                });
+        }
+
+        private void onGameInitializedCore(Game game)
+        {
+            mapper.Reset();
+            currentGame = game;
+            var credentials = getCredentials(game);
```
njthomson: `credentials` feels like a good thing to cache - it shouldn't be possible to get a different value for the duration of the session. This way we A) save recomputing below on line 132 and B) can disable things outright if we are unable to get a good value at this point.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+        private static InaraCredentials? getCredentials(Game game)
+        {
+            var commander = resolveCommanderName(game.cmdr?.inaraCommanderName, game.Commander);
+            var frontierId = game.fid;
+            var apiKey = game.cmdr?.inaraApiKey?.Trim();
+            if (string.IsNullOrWhiteSpace(commander) || string.IsNullOrWhiteSpace(apiKey)) return null;
+            return new InaraCredentials(commander, frontierId ?? string.Empty, apiKey);
+        }
+
+        internal static string? resolveCommanderName(string? configuredName, string? journalName)
```
njthomson: This function is used only once above. It doesn't do much heavy lifting, so I would just inline this code above.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+            game.journals?.Entries.OfType<Fileheader>().FirstOrDefault()?.gameversion
+            ?? game.journals?.Entries.OfType<LoadGame>().LastOrDefault()?.gameversion;
+
+        private static InaraCredentials? getCredentials(Game game)
+        {
+            var commander = resolveCommanderName(game.cmdr?.inaraCommanderName, game.Commander);
+            var frontierId = game.fid;
+            var apiKey = game.cmdr?.inaraApiKey?.Trim();
+            if (string.IsNullOrWhiteSpace(commander) || string.IsNullOrWhiteSpace(apiKey)) return null;
+            return new InaraCredentials(commander, frontierId ?? string.Empty, apiKey);
```
njthomson: > ... frontierId ?? string.Empty

Wouldn't it be a serious error condition if we have no fid value? I would throw an exception right here, rather than waiting for problems to happen down the line.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+            if (string.IsNullOrWhiteSpace(gameVersion)) return false;
+
+            var numeric = new string(gameVersion.TakeWhile(character => char.IsDigit(character) || character == '.').ToArray());
+            return Version.TryParse(numeric.TrimEnd('.'), out var version) && version.Major >= 4;
+        }
+
+        internal async Task flushAsync() => await sendPendingAsync();
+
+        private static string? getGameVersion(Game game) =>
+            game.journals?.Entries.OfType<Fileheader>().FirstOrDefault()?.gameversion
```
njthomson: I believe what you have forces code to process all entries looking for any matches, then you pick the first. Consider something like, which stops as soon as it finds the 1st match:

`(game?.journals?.Entries.FirstOrDefault(x => x is Fileheader) as Fileheader)?.gameversion`

Or better yet, maybe we cache it in _Journal.cs_, where we cache if `isOdyssey`, see: https://github.com/njthomson/SrvSurvey/blob/b9cac22183f00d846fbbaca4c47a40d1677532c4/SrvSurvey/game/JournalFile.cs#L58

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+            ?? game.journals?.Entries.OfType<LoadGame>().LastOrDefault()?.gameversion;
```
njthomson: I would assume there's **always** a `Fileheader`, hence this is redundant, no?

SrvSurvey/net/Inara.cs
njthomson: In fact, looking over other code I have, I've made an assumption that it is always the first line in a journal file. *(and I've yet to experience a problem from that)*

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+            if (!ReferenceEquals(currentGame, game))
+            {
+                mapper.Reset();
+                currentGame = game;
+            }
+
+            var credentials = getCredentials(game);
+            var gameVersion = getGameVersion(game);
+            var isLive = IsLiveVersion(gameVersion, game.journals?.isOdyssey == true);
+            var isBeta = IsBetaVersion(gameVersion);
```
njthomson: Similar to my other comment, I would not expect these 4 values could change withing a session, so it makes sense to cache the values up front.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+            allowSharedStatus ? game.status?.InTaxi : null);
+
+        private static bool isMultiboxing() => DetectMultiboxing(
+            Elite.hadManyGameProcs,
+            countGameProcesses,
+            ex => Game.log($"Inara could not count Elite processes and conservatively enabled multi-box suppression ({ex.GetType().Name})."));
+
+        private static int countGameProcesses()
+        {
+            var gameProcesses = Elite.GetGameProcs();
```
njthomson: This will be a problem. I used to call this often and it would use a lot of CPU - people complained a lot. *(Powerful machines do okay, but older ones suffered badly)*

Use `Elite.hadManyGameProcs` which is efficiently maintained by a separate timer and safe to call with frequency.

njthomson: In fact, I'm already tracking if multi-boxing is happening elsewhere. You could use `FormMultiFloatie.current != null` to know when multi-boxing is happening. I would use that and remove `DetectMultiboxing` + related code, leaving just:

``` C#
private static bool isMultiboxing() => FormMultiFloatie.current != null;
```

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+            }
+            catch (Exception ex)
+            {
+                try { onError?.Invoke(ex); }
+                catch { /* optional diagnostics must not escape */ }
+                return true;
+            }
+        }
+
+        internal static bool RunIsolated(Action action, Action<Exception>? onError = null)
```
njthomson: > internal ...
This *(and many static functions below)* should be private ... unless there's a hard reason to expose them to external code?

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+            }
+            finally
+            {
+                Interlocked.Exchange(ref sending, 0);
+            }
+        }
+
+        private static bool isTransient(HttpStatusCode status) =>
+            status == HttpStatusCode.RequestTimeout
+            || status == HttpStatusCode.TooManyRequests
```
njthomson, line 491: I remember Inara doesn't like it if people are uploading from multiple clients at the same time ... I wonder if we want to pop a message or alert the user if we start seeing this?

*(Or add a TODO for this to be done in a future PR)*

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+            var uploadClient = client;
+            if (uploadClient == null) return;
+
+            if (Interlocked.Exchange(ref sending, 1) != 0) return;
+            try
+            {
+                var pending = queue.TakeAll();
+                if (pending.Count == 0) return;
+
+                foreach (var group in pending.GroupBy(item => item.Credentials))
```
njthomson: > GroupBy(item => item.Credentials))

Just curious ... within a game session, would entries ever have different credentials?

*(I'm assuming we should also flush any pending events when a `Game` instance is shutting down, to ensure no mixed events. Even then though, it would be the same player - there is no log-out/log-in without killing the game process)*

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+                {
+                    var batch = group.ToList();
+                    if (!Game.settings.inaraUpload)
```
njthomson: Why not have this up front and return early if the setting is disabled?

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+                        continue;
+
+                    var activeGame = Game.activeGame;
+                    if (activeGame != null
+                        && string.Equals(activeGame.Commander, group.Key.Commander, StringComparison.OrdinalIgnoreCase)
+                        && !string.Equals(activeGame.cmdr?.inaraApiKey?.Trim(), group.Key.ApiKey, StringComparison.Ordinal))
+                    {
+                        Game.log($"Inara discarded {batch.Count} queued event(s) after the commander API key changed.");
```
njthomson: Could there be a race hazard, where the `Game.activeGame` instance *just* got cleared and we still have some entries to send? It would be a shame to deny these from Inara.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+                        if (isTransient(response.StatusCode))
+                        {
+                            queue.Requeue(batch);
+                            Game.log($"Inara upload deferred after HTTP {(int)response.StatusCode}; {batch.Count} event(s) retained.");
+                            continue;
+                        }
+
+                        if (!response.IsSuccessStatusCode)
+                        {
+                            Game.log($"Inara rejected {batch.Count} event(s) with HTTP {(int)response.StatusCode}.");
```
njthomson: Add `response.ReasonPhrase` here - it may have something useful for diagnostics.
*(Same for line 429)*

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+                            continue;
+                        }
+
+                        var body = await response.Content.ReadAsStringAsync();
+                        if (string.IsNullOrWhiteSpace(body))
```
njthomson: Just curious ... does Artie want us to re-send events in this case? Is this an error case, or does it happen often?

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+                        if (!responseIsComplete)
+                        {
+                            queue.Requeue(batch);
+                            Game.log($"Inara returned an incomplete response; {batch.Count} event(s) retained.");
+                            continue;
+                        }
+
+                        if (headerStatus is >= 400)
+                        {
+                            Game.log($"Inara rejected a batch of {batch.Count} event(s) with API status {headerStatus}.");
```
njthomson: Have you seen this in a debugger? Is there anything else we can log for diagnostics?

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+                        if (string.IsNullOrWhiteSpace(body))
+                        {
+                            queue.Requeue(batch);
+                            Game.log($"Inara returned an empty response; {batch.Count} event(s) retained.");
+                            continue;
+                        }
+
+                        var result = JObject.Parse(body);
+                        var headerStatus = result.SelectToken("header.eventStatus")?.Value<int?>();
+                        var responseEvents = result["events"] as JArray;
```
njthomson: To avoid null checks below, consider:
`var responseEvents = result["events"] as JArray ?? [];`

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+                            continue;
+                        }
+
+                        if (headerStatus is >= 400)
+                        {
+                            Game.log($"Inara rejected a batch of {batch.Count} event(s) with API status {headerStatus}.");
+                            continue;
+                        }
+
+                        var failedEvents = responseEvents!.OfType<JObject>()
```
njthomson: > OfType
Would they be anything else? There's a way to cast everything in a collection, which would be cheaper at runtime when needing to filter the list.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+
+            return raw;
+        }
+
+        private async Task sendPendingAsync()
+        {
+            var uploadClient = client;
+            if (uploadClient == null) return;
+
+            if (Interlocked.Exchange(ref sending, 1) != 0) return;
```
njthomson: You're using this as a bool. And, as far as I remember, reading and writing boolean members is naturally atomic and thread safe, so the use of `Interlocked.Exchange` is overkill.

*(I did not know about `Interlocked.Exchange` - that is a handy thing nonetheless)*

SrvSurvey/net/Inara.cs
Fenris159: Yeah I haven't explored enough with multiboxing to safely understand how to provide network features safely so my solution was to just disable some of them until the session is solo again. 

For Inara specifically it continues safe journal-native uploads and suppresses ambiguous shared sidecars/status. This was "I felt" a good compromise until another pr could be built around safer Network features.

SrvSurvey/net/Inara.cs
Fenris159: That's a good catch. I meant to put a disclaimer on the Opt-in window advising them to dedicate one application for Inara and not use it with EDMC enabled tracking on top of it. I thought about maybe even a process detection to throw a warning but I thought maybe too aggressive.

SrvSurvey/Settings.cs
Fenris159: Inara says development clients should explicitly send `true` and should not rely on the server-side override, so known testers should manually enable the hidden setting. Normal users should receive `false`. https://inara.cz/elite/inara-api-docs/

I could default the permanent property to `false`. If fresh test installations truly must default on temporarily, I'll use your temporary `inaraDeveloperTestModeTEST` property or we can an explicit one-time migration.

SrvSurvey/net/InaraEventMapper.cs
Fenris159: I can add commit-pinned links to each derived file based on EDMC commit `99709c0` in case branch has changes later on. "Although they are very slow to making changes like that"

SrvSurvey/net/Inara.cs
njthomson: Thinking some more ... I accept that `Interlocked.Exchange` is still be beneficial as we need to change the value and check it, but ... are we supposed to hit this point every ~20 seconds? Might it happen as a real close race hazard?

SrvSurvey/net/Inara.cs
Fenris159: I could remove `currentGame` entirely. `onGameInitialized` already resets the mapper. If necessary also add an explicit session-ended callback for flushing/cleanup.

SrvSurvey/net/Inara.cs
Fenris159: - 20 seconds: maximum duration of one HTTP request.
- 35 seconds: minimum batching cadence.

Those values match EDMC’s pinned implementation: [`[_TIMEOUT = 20](https://github.com/EDCD/EDMarketConnector/blob/99709c0146e0e64e364ebc744300c184a2e6d3c9/plugins/inara.py#L54)`](https://github.com/EDCD/EDMarketConnector/blob/99709c0146e0e64e364ebc744300c184a2e6d3c9/plugins/inara.py#L54) and [`[WORKER_WAIT_TIME = 35](https://github.com/EDCD/EDMarketConnector/blob/99709c0146e0e64e364ebc744300c184a2e6d3c9/plugins/inara.py#L154-L155)`](https://github.com/EDCD/EDMarketConnector/blob/99709c0146e0e64e364ebc744300c184a2e6d3c9/plugins/inara.py#L154-L155).

I decided to mirror their configuration as they have been updating Inara for a very long time without complaints so I drew assumptions based on that rather than asking Artie directly.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+            catch (Exception ex)
+            {
+                configuredTimer?.Dispose();
+                configuredClient?.Dispose();
+                configuredTimer = null;
+                configuredClient = null;
+                RunIsolated(() => Game.log($"Inara initialization was disabled without affecting SrvSurvey ({ex.GetType().Name})."));
+            }
+
+            client = configuredClient;
```
njthomson: The reality is that, unless something catastrophic happens, `this.client` should always be populated - but you have to treat it as _potentially_ null everywhere below.

May I suggest using a static factory method that does what you have in the constructor here, but it either returns a valid instance of `Inara` or null should those catastrophic things happen. This way you can guarantee that `timer` and `client` are always populated but still handle error cases, and we never end up with an undead instance of `Inara` floating around.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+            timer = configuredTimer;
+        }
+
+        public void Dispose()
+        {
+            RunIsolated(() => timer?.Dispose());
```
njthomson: Consider:
```
RunIsolated(() => {
	timer?.Dispose();
	client?.Dispose();
});
```

Though to be honest, this feels overkill too as things throwing in their Dispose method is exceptionally rare *(things implemented in the .Net framework that is)*

SrvSurvey/net/Inara.cs
njthomson: Thinking further, I strongly think we should not hold a reference to `Game` from a static singleton.

SrvSurvey/net/Inara.cs
Fenris159: I can include the call-stack, no problem with that.

SrvSurvey/net/Inara.cs
Fenris159: The identity check is safer than merely testing `activeGame == null`. It prevents callbacks from stale or disposing `Game` instances.

`!inaraUpload` return would break mid-session behavior. The mapper intentionally stays warm while uploads are disabled.

I would just keep it as is.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+            RunIsolated(() => client?.Dispose());
+        }
+
+        public void onGameInitialized(Game game)
+        {
+            if (client == null) return;
+
+            RunIsolated(
```
njthomson: (nit) I'm not convinced that `RunIsolated` provides enough benefit to really exist. No code that calls it takes advantage of the bool it returns. The following is equivalent, but it's less obfuscated and reduces the number of stack frames:
``` C#
try
{
  onGameInitializedCore(game);
}
catch(exception ex)
{
  mapper.Reset();
  currentGame = game;
  Game.log($"Inara startup seeding was skipped without affecting SrvSurvey ({ex.GetType().Name}).");
}
```

I would also strongly recommend including the exception message and stack-trace in logging.
*(Along with any other place you trace from a `catch` block in this file)*

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+                configuredClient.DefaultRequestHeaders.Add("user-agent", Program.userAgent);
+                configuredTimer = new System.Threading.Timer(_ => sendPendingAsync().justDoIt(), null, sendInterval, sendInterval);
+            }
+            catch (Exception ex)
+            {
+                configuredTimer?.Dispose();
+                configuredClient?.Dispose();
+                configuredTimer = null;
+                configuredClient = null;
+                RunIsolated(() => Game.log($"Inara initialization was disabled without affecting SrvSurvey ({ex.GetType().Name})."));
```
njthomson: Use of `RunIsolated` here feels overkill.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+            var credentials = getCredentials(game);
+            var canPrepareUpload = CanPrepareUpload(
+                Game.settings.inaraUpload,
+                credentials?.ApiKey,
+                IsLiveVersion(getGameVersion(game), game.journals?.isOdyssey == true),
+                IsBetaVersion(getGameVersion(game)));
+            var multiboxing = canPrepareUpload && isMultiboxing();
+
+            var filepath = game.journals?.filepath;
+            if (string.IsNullOrWhiteSpace(filepath)) return;
```
njthomson: *(nit)* I recommend doing this first, as there's little point with any of the code above if we have no `filepath`.

SrvSurvey/net/Inara.cs
Fenris159: I figured the current recomputation cost was negligible. If you feel like it's necessary we could cache the credentials but the Inara settings window can change the API key and upload opt-in during the session. We might have to still continue reading the API key and opt-in state when processing or sending or otherwise rebuild the cache on changes but that seemed like extra work.

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+                if (string.IsNullOrWhiteSpace(line)) continue;
+
+                JObject entry;
+                try
+                {
+                    entry = JObject.Parse(line);
+                }
+                catch (JsonException)
+                {
+                    continue;
```
njthomson: I recommend tracing something

SrvSurvey/net/Inara.cs
Fenris159: It exists mainly as a focused test seam. I can inline it and test through credential resolution.

SrvSurvey/net/Inara.cs
Fenris159: No it should be fine. Inara explicitly says `commanderFrontierID` should be omitted when unknown. The payload builder already omits an empty value. [Inara API documentation](https://inara.cz/elite/inara-api-docs/)

SrvSurvey/net/Inara.cs
```diff
@@ -0,0 +1,501 @@
...
+using SrvSurvey.game;
+using System.Net;
+using System.Text;
+
+namespace SrvSurvey.net
+{
+    /// <summary>
+    /// Collects and batches opted-in commander updates for Inara.
+    /// </summary>
+    internal sealed class Inara : IDisposable
```
njthomson, line 16: Having been over this file a few times now, I have a design suggestion...

Rather than being a long-lived static singleton, if you create an instance of `Inara` as a regular instance member of `Game` - you *could* keep a reference to a `Game` instance and **not** have to rely on `Game.activeGame` *(which is safer)*. You would need to make sure `Game.Dispose` disposes the instance of `Inara`, but ... this helps track a set of events with guaranteed identical credentials/context, and you can cleanly handle the final processing of events internally as `Inara` is disposed *(or something like that)*

SrvSurvey/net/Inara.cs
Fenris159: What if we cache `gameversion` once in `JournalFile`, next to `isOdyssey`, using the first `Fileheader` and retaining a defensive `LoadGame` fallback? Then we can avoid unnecessarily retrieving a fixed game version repeatedly on every journal event.

SrvSurvey/game/CommanderSettings.cs
```diff
@@ -141,8 +141,12 @@ private static void seedCommanderSettings()
         public string fid;
         public string commander;
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore, DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
-        public string? rccApiKey;
-        public bool isOdyssey;
+        public string? rccApiKey;
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore, DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        public string? inaraApiKey;
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore, DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        public string? inaraCommanderName;
```
njthomson: For my education ... how does this differ from the regular Commander name? Does Inara allow them to be different?

SrvSurvey/net/Inara.cs
Fenris159: Seems valid for a normal initialized game session.
The `LoadGame` fallback is defensive but probably unnecessary.
If `gameversion` is cached from the first `Fileheader`, removing that fallback is reasonable.

SrvSurvey/forms/FormInaraIntegration.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            if (this.checkUpload.Checked && string.IsNullOrWhiteSpace(this.txtApiKey.Text))
+            {
+                MessageBox.Show(this, "Enter your Inara API key before enabling uploads.", "Inara Integration", MessageBoxButtons.OK, MessageBoxIcon.Warning);
+                this.txtApiKey.Focus();
+                return;
+            }
+
+            Game.settings.inaraUpload = this.checkUpload.Checked;
+            Game.settings.Save();
```
njthomson: Sadly we must not do this here. The user may have other settings changes pending from FormSettings, if they hit your Save button then Cancel from FormSettings ... we would have committed all changes here unbeknownst to the the user, despite the fact they hit Cancel.

We can save changes to `this.cmdrSettings` but not `Game.settings`. But ... you can change values in `Game.settings` and wait for the Save button in the parent form to commit this. 

From a pure semantics perspective ... if we change the button text from `Save` to `OK` it is might help a little?

SrvSurvey/net/Inara.cs
Fenris159: The API key and global opt-in are editable mid-session, but your right about the immutable values.

SrvSurvey/forms/FormInaraIntegration.cs
njthomson: I wonder ... we can use the presence of an inara api-key instead of having setting `Game.settings.inaraUpload` at all? Meaning, a user elects to upload to Inara by setting the api-key, and if they want to stop, they come back to this dialog and clear it.

*(Naturally, we should probably add a button on this window to explicitly clear the api-key)*

This also helps people with many alts. Having `Game.settings.inaraUpload = true` but no api-key means no uploading to Inara is going to happen.

Effectively, in _Game.cs_ `Journals_onJournalEntry` you'd have:
``` c#
if (!string.IsNullOrEmpty(this.cmdr.inaraApiKey))
  inara.onJournalEntry(this, raw);
```

SrvSurvey/net/Inara.cs
Fenris159: I double checked this one with AI. It was suggesting we use `Elite.hadManyGameProcs`, which is refreshed every five seconds.

Apparently `FormMultiFloatie.current` is not reliable? It remains null when the user hides the multi-floatie even though multiple processes are running, or at least that's what it's telling me.

Fenris159: Then we just remove Inara’s `GetGameProcs`, `countGameProcesses`, and `DetectMultiboxing` path.

SrvSurvey/forms/FormInaraIntegration.cs
```diff
@@ -0,0 +1,107 @@
...
+            this.checkUpload.Checked = Game.settings.inaraUpload;
+
+            if (this.cmdrSettings != null)
+            {
+                this.txtCommander.Text = string.IsNullOrWhiteSpace(this.cmdrSettings.inaraCommanderName)
+                    ? this.cmdrSettings.commander
+                    : this.cmdrSettings.inaraCommanderName;
+                this.txtApiKey.Text = this.cmdrSettings.inaraApiKey ?? string.Empty;
+            }
+            else
```
njthomson, line 33: This error case is a bit dubious ... we cannot save anything, and there's no point really in showing the dialog.

*(And I accept that I do not handle the equivalent case when setting the RCC api-key at all - this is something I should address)*

I'll put something in once this PR is done. *(You should leave it be for now, or if anything, add a TODO comment)*

SrvSurvey/forms/FormSettings.cs
```diff
@@ -944,11 +943,6 @@ private void checkBox49_CheckedChanged(object sender, EventArgs e)
             checkBox48.Enabled = checkColonization.Checked && checkBox49.Checked && checkBox43.Checked;
         }
 
-        private void linkLabel3_LinkClicked(object sender, LinkLabelLinkClickedEventArgs e)
```
njthomson, line 947: *(thank you for this)*

SrvSurvey/Settings.cs
njthomson: We could literally hard-code it where `inaraDeveloperTestMode` is read from and remove that setting for now. We then re-add the setting once we're ready for it to be not hard-coded. *(defaulting to `false` at that time)*

SrvSurvey/net/Inara.cs
njthomson: Let's leave it be then. If it works for EDMC ... we should be no different.

SrvSurvey/net/Inara.cs
njthomson: I concur the processing is negligible, but handling a change in settings is valid. Let's leave it be for now. *(Especially if EDMC has the same approach)*

SrvSurvey/net/Inara.cs
njthomson: Perhaps, but as far as SrvSurvey is concerned ... it should *always* know the FID. If not ... I think that would be a weird error condition.

The only case it does not know is when the game just started, is in the main menu still and the user didn't start playing yet. Once they start playing ... from that point forwards, the FID value is always known.

SrvSurvey/net/Inara.cs
njthomson: > Apparently FormMultiFloatie.current is not reliable? It remains null when the user hides the multi-floatie even though multiple processes are running, or at least that's what it's telling me.

Yeah, you're right about that ... using just `Elite.hadManyGameProcs` should be good then,

SrvSurvey/net/Inara.cs
njthomson: I rescind this comment, understanding that test code needs it.

SrvSurvey/net/Inara.cs
Fenris159: A `SemaphoreSlim` or tracked active-send task could be the cleaner session-shutdown mechanism to use?

SrvSurvey/net/Inara.cs
njthomson: >  I thought about maybe even a process detection to throw a warning but I thought maybe too aggressive.

I concur. Just because the other processes may be running, they may have adjust settings to not upload to Inara.

SrvSurvey/net/Inara.cs
njthomson: I get it now ... people changing their Inara api-key in the middle of a session is relevant.

SrvSurvey/net/Inara.cs
Fenris159:
```
if (string.IsNullOrWhiteSpace(game.fid))
{
    Game.log("Inara was not created because the active commander has no FID.");
    return null;
}

```
This could reconcile both:
- Inara’s API technically permits an omitted FID.
- SrvSurvey should treat a missing FID after game initialization as abnormal.

SrvSurvey/game/CommanderSettings.cs
Fenris159: Inara’s API documentation says `commanderName` should be the in-game name obtained from the journal or cAPI—not a user-entered Inara name. The personal API key identifies the Inara account.

So these should not differ for our purposes. I’ll use the regular journal-derived `Game.Commander` value and remove `inaraCommanderName`, its UI field, and the related resolution helper.

I think that happened when I initially had the AI mock up the forms and it treated that as a user set entry form.

njthomson: Use the name from `CommanderSettings` instead. I don't *think* they would be different, but there are some potential edge cases when people change settings and the game isn't running. Or maybe leave the settings dialog open, shut down the game and restart it as another alt. *(not a likely thing, but you never know)*

Fenris159:
For the sake of time I'm going to turn the remaining comments into a spec by having my agent acknowledge each of your remaining findings and provide a proper response then rebuild the RP based on the replies so it can track with the redesign. I decided to follow your advice.

> Having been over this file a few times now, I have a design suggestion...
> 
> Rather than being a long-lived static singleton, if you create an instance of Inara as a regular instance member of Game - you could keep a reference to a Game instance and not have to rely on Game.activeGame (which is safer). You would need to make sure Game.Dispose disposes the instance of Inara, but ... this helps track a set of events with guaranteed identical credentials/context, and you can cleanly handle the final processing of events internally as Inara is disposed (or something like that)

SrvSurvey/Settings.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid, with the later hard-coding proposal providing the cleanest transition. During the application-validation phase, build `isBeingDeveloped = true` directly into payload construction and remove the persisted setting so test values cannot become sticky. Reintroduce a hidden setting only when production mode is enabled, defaulting it to `false`, and update the PR description to match.

SrvSurvey/net/InaraEventMapper.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid. Each source file materially derived from EDMC should cite an immutable commit permalink rather than `main`; this preserves the exact behavioral reference through later upstream changes. The rewrite spec uses the EDMC commit contemporaneous with this implementation and keeps the GPL attribution beside the derived logic.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid, and superseded by the later per-`Game` ownership design. `Inara` should be a regular member of one `Game`, may safely retain that owning `Game`, and must be stopped/disposed when that `Game` is disposed. This removes the static lifetime mismatch, `currentGame`, and all reliance on `Game.activeGame` for upload identity.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Resolved by the later agreement to mirror EDMC. The 20-second value is the maximum duration of one HTTP request; 35 seconds is the batching/send cadence, so they intentionally differ. The recommended rewrite leaves both as constants.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid future work, but it should remain outside this rewrite. An exact `Cargo.json`/journal `Cargo` timestamp match can make that one ship snapshot attributable during multiboxing; it does not establish ownership of `Status.json` or `ShipLocker.json`. Keep the present conservative suppression and cover timestamp correlation in a separate PR.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid. Catch blocks should log the exception message and call stack, preferably by logging the exception object, while never logging request JSON, API keys, or other credentials. The same diagnostic standard should be applied consistently across the Inara module.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Superseded by the per-`Game` architecture. A session-owned Inara receives entries only from its owning `Game`, so it should not accept a `Game` parameter or inspect `Game.activeGame`. It should still process live entries while that commander's API key is absent so mapper state remains warm for mid-session enablement; only event collection/upload is gated by the key.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** The later agreement to leave inexpensive credential recalculation is sound. Cache immutable session identity (commander, FID, game version, and Live/beta state), but read the API key from the owning session's exact `CommanderSettings` so a key edit takes effect safely. Queued events retain the credential snapshot under which they were created.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** The one-use helper should disappear, but primarily because the editable `inaraCommanderName` override should be removed. Inara's `commanderName` must come from the session's `CommanderSettings.commander`; the personal API key identifies the Inara account. With one authoritative name source, there is no resolution helper to inline.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid as a SrvSurvey session invariant even though Inara technically permits an omitted FID. A per-`Game` Inara is created only after commander initialization, so a missing FID indicates an invalid session. The factory should log the full diagnostic and return `null`, preventing Inara startup without allowing the error to disrupt normal SrvSurvey processing.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** The caching outcome is valid, although `OfType<Fileheader>().FirstOrDefault()` itself is lazy and does not enumerate past the first match. Under the per-`Game` design, resolve the version once from the first `Fileheader` during Inara creation and cache the resulting Live/beta eligibility for the session.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid for an initialized `Game`. `Game.initializeFromJournal` already treats the first entry as `Fileheader` before Inara can be created, so the `LoadGame` fallback is unnecessary in the session-owned implementation. Use the established first-entry invariant and fail Inara creation cleanly if it is unexpectedly violated.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Partly valid. Cache the session's commander, FID, game version, and Live/beta classification in the per-`Game` Inara. Do not freeze the API key: it belongs to that same session `CommanderSettings` object but may be edited mid-session, so reads should remain dynamic and old-key queued groups must be discarded after a change.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid, with the later conclusion confirmed. Use only `Elite.hadManyGameProcs`, which is refreshed by the existing five-second process monitor. Remove Inara's direct `GetGameProcs`, `countGameProcesses`, and `DetectMultiboxing` path; `FormMultiFloatie.current` is not reliable when that UI is hidden.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Resolved by the rescission. The focused `internal` methods are intentional test seams exposed to `SrvSurvey.Tests` through `InternalsVisibleTo`, not public application interfaces. They should remain internal where direct regression testing is valuable; methods made obsolete by the rewrite can simply be removed.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid as user guidance, with the later rejection of process detection also correct. Add concise text to the Inara settings window telling users to enable Inara uploads in only one application at a time. Do not infer EDMC's upload setting merely from its process being present.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** The later observation about mid-session API-key changes is correct. Per-`Game` ownership prevents different commanders from sharing a queue, while credential grouping remains useful when the same commander's key changes during that session. Each queued event retains its original credential snapshot; groups whose key no longer matches that session's current key are discarded.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** This becomes obsolete if API-key presence is the per-commander opt-in and the global `inaraUpload` setting is removed. The mapper should continue warming even without a key. At send time, an absent or changed key must drain/discard incompatible queued events rather than return while leaving them eligible for a later accidental upload.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Superseded by session ownership. Final processing must use the Inara instance's bound commander/FID and that same `Game.cmdr` API key; it should never consult `Game.activeGame`. Switching commanders disposes and finalizes the old instance before constructing the new session instance, eliminating this cross-session race.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid. Include both the numeric HTTP status and `ReasonPhrase` on transient and permanent HTTP failures. Keep request bodies and credential-bearing headers out of diagnostics.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Re-sending is correct here: Inara's documentation says an empty response means something went wrong and the request should be tried again after a reasonable delay. Retain the batch with bounded backoff/jitter rather than retrying aggressively, and ensure session disposal does not lose that retained batch.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid. No live Inara write/debugger observation is recorded for this PR, so diagnostics should include `header.eventStatus`, safe/truncated `header.eventStatusText`, and each rejected event's name/status/text. Do not log the outbound payload because its header contains the personal API key.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** The null-coalescing suggestion should not be applied. A missing `events` property is a malformed/incomplete response, not an empty successful response. Preserve that distinction and requeue the batch through the incomplete-response path.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Partly valid. Direct `Cast<JObject>()` is appropriate only after validating that every response token is an object containing `eventStatus`; otherwise it throws or a filtering approach silently ignores malformed entries. Validate the complete shape first, then cast and process it.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** The later acceptance of an atomic gate is correct. The periodic timer, a journal `Shutdown`, and per-`Game` disposal can overlap even though the normal cadence is 35 seconds and 20 seconds is only the request timeout. A `SemaphoreSlim` or tracked active-send task is preferable for shutdown because disposal must wait for an active send and then perform exactly one final flush rather than merely skip it.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid and complementary to per-`Game` ownership. Use a static creation method that receives the initialized `Game` and returns either a fully usable Inara with non-null client/timer/session identity or `null` after logging. This removes the nullable-client checks and prevents a partially initialized instance from surviving.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid in principle: framework timer/client disposal does not justify broad isolation wrappers. The rewrite should stop the timer, serialize/finish any active or final send, and then dispose the HTTP resources directly. It must not start a fire-and-forget flush and immediately dispose the client that flush needs.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid. Remove `RunIsolated` and use local `try/catch` blocks where the recovery policy is visible at the call site; its unused Boolean result adds no leverage. Every caught exception should log its message and stack trace without exposing credentials.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid. The per-`Game` factory should contain one explicit initialization `try/catch`, return `null` on failure, and clean up any partially created resources locally. Calling `RunIsolated` solely to write an initialization log is unnecessary.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid. Check the journal filepath and the initialized session identity before resetting or replaying mapper state. In the revised design this validation belongs in the per-`Game` creation/seeding path, before any timer is started.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid, but avoid logging the malformed journal line itself because it may contain commander data. Count skipped malformed lines during replay and emit one aggregate diagnostic such as `Inara skipped N malformed journal entries while seeding`.

SrvSurvey/net/Inara.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid and adopted as the governing rewrite architecture. Each `Game` owns exactly one Inara session module, created after commander initialization and disposed before that Game's context is torn down. The module binds to `game.cmdr.commander`, `game.cmdr.fid`, and that same `game.cmdr.inaraApiKey`; switching commanders disposes the old module and creates a new one, so names, keys, mapper state, queues, and final events cannot cross sessions.

SrvSurvey/game/CommanderSettings.cs
Fenris159: **Codex review-agent assessment (not the PR author):** The regular name from the session's `CommanderSettings.commander` is the correct value. Inara's `commanderName` means the in-game CMDR name; the personal API key identifies the Inara account. Remove `inaraCommanderName` and its editable control, bind both name and key to the exact `CommanderSettings` selected by the session FID, and keep the dialog bound to that captured settings object even if the active game changes while it is open.

SrvSurvey/forms/FormInaraIntegration.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Both findings are valid. The child dialog may save only the captured per-commander `CommanderSettings`; it must not call `Game.settings.Save()` and thereby commit unrelated parent edits. Remove the global `inaraUpload` flag, use the presence of that commander's API key as opt-in, label the confirmation button `OK`, and provide an explicit Clear Key action. The per-`Game` Inara should still receive live events without a key so state remains warm; collection/upload starts only when its own commander's key is present.

SrvSurvey/forms/FormInaraIntegration.cs
Fenris159: **Codex review-agent assessment (not the PR author):** Valid and explicitly deferred by the owner. Leave the current behavior unchanged in this rewrite, optionally recording the requested TODO. The settings entry point can be hardened consistently with the Raven key dialog in a later change.

SrvSurvey/game/Game.cs
```diff
@@ -251,16 +250,18 @@ public void Dispose()
...
                 EDDN.header = null;
 
-                if (this.journals != null)
-                {
-                    this.journals.onJournalEntry -= Journals_onJournalEntry;
-                    this.journals.Dispose();
+                if (this.journals != null)
+                {
+                    this.journals.onJournalEntry -= Journals_onJournalEntry;
+                    this.inara?.Dispose();
```
njthomson: *(nit)* I would put these two lines outside of the IF block. We want to dispose `this.inara` even if there happens to be no `this.journals`

SrvSurvey/game/Game.cs
```diff
@@ -278,9 +279,15 @@ protected virtual void Dispose(bool disposing)
...
             }
 
-            if (Game.activeGame == this)
-                Game.activeGame = null;
-        }
+            if (Game.activeGame == this)
+                Game.activeGame = null;
+        }
+
+        internal void onInaraApiKeyChanged(CommanderSettings settings)
```
njthomson: It feels reasonable to me that `this.inara` be made public, then the dialog can access it directly and there's no need for this function. *(I like the idea of `Game` being largely unaware of Inara stuff)*

SrvSurvey/net/Inara.cs
```diff
@@ -4,91 +4,119 @@
...
     /// <summary>
-    /// Collects and batches opted-in commander updates for Inara.
+    /// Collects and batches Inara updates for exactly one initialized game session.
     /// </summary>
     internal sealed class Inara : IDisposable
     {
         internal const string Endpoint = "https://inara.cz/inapi/v1/";
         private static readonly TimeSpan sendInterval = TimeSpan.FromSeconds(35);
-        private readonly HttpClient? client;
+        private readonly Game? game;
```
njthomson: We should assume `this.game` is always populated. Meaning: the factory create method can return null if it is missing.

Hence, we shouldn't need to do `this.game!.xxx` anywhere below.

SrvSurvey/net/Inara.cs
```diff
@@ -389,113 +357,220 @@ private static JObject addSidecarData(Game game, JObject raw, bool allowSharedSi
...
+            }
+        }
+
+        public void Dispose()
+        {
+            try
+            {
+                StopAsync().GetAwaiter().GetResult();
+            }
+            catch (Exception ex)
```
njthomson, line 551: If there's no `await` here, so ... we'll probably leave this catch block before `StopAsync()` etc finishes. Only ... we cannot have `await` in a `Dispose` method, so we should have some other mechanism to handle exceptions thrown asynchronously.

njthomson:
Nice work @Fenris159

PR #1026, opened 2026-07-28
Complete opt-in EDDN synchronization

Fenris159:
## Summary

- Rebase the EDDN integration onto current `main` and replace the long-lived static uploader with two explicit lifetimes:
  - `Main` owns durable delivery, transport, consent checks, and the outbox.
  - Each `Game` owns an `EddnSessionPublisher` with an immutable Commander header and session state.
- Add a dedicated **Configure EDDN Sharing** dialog immediately above the Inara button. Sharing remains off until the user explicitly opts in.
- Include the journal/companion-file disclosure, local retry-queue behavior, Commander uploader identity, multi-client safety behavior, and the warning to enable EDDN in only one uploader such as SrvSurvey or EDMC.
- Use EDDN's Live gateway while appending `/test` to schema references internally. This is intentionally not a user setting and can be changed by a future release.
- Publish supported raw journal events and Market, Outfitting, Shipyard, Fleet Carrier Materials, and NavRoute companion data using the filtering and transformation patterns documented by EDDN and EDMarketConnector.

## Session and delivery safety

- Construct EDDN only after `Settings.Load()` has completed, avoiding static-startup access to an uninitialized `Settings.instance`.
- Capture the Commander name, game version, and SrvSurvey version in each session. Switching commanders disposes the old publisher, so a later session cannot reuse the previous Commander's identity.
- Persist immutable envelopes with their captured header. The process-owned outbox can finish an older session's queued messages without consulting the current `Game` or Commander.
- Capture FSS signal-batch location and expansion flags when the batch begins, so a jump cannot relabel the previous system's signals.
- Invalidate delayed companion-file reads on session changes or opt-out.
- Pause publication while multiple Elite clients make shared Status/companion files ambiguous, while retaining pending messages for a safe retry.

## Durable outbox

- Persist one atomic file per queued message instead of rewriting the complete queue on every journal event.
- Preserve first-attempt order while allowing later due messages to proceed when an earlier message is waiting for retry.
- Quarantine individual corrupt queue entries, bound the backlog, use an exclusive cross-process lease, and clear pending uploads when sharing is disabled.
- Gzip requests, use HTTP/1.1, classify transient and permanent failures, and recheck consent and process safety before network delivery.

## Privacy and authentication

- EDDN requires no account, API key, OAuth flow, or authorization header.
- The current session's Commander name is sent as EDDN's required `uploaderID`; EDDN obfuscates it before redistribution.
- Only newly observed live events are considered. SrvSurvey does not replay journal history to EDDN.
- Schema-disallowed localized and Commander-specific fields are removed before persistence.

## References

- EDMarketConnector EDDN implementation: `EDCD/EDMarketConnector@2b6a0ce1ee3ba60c21f3f4e9fa093046da8825e4`
- EDDN schemas and uploader guidance: `EDCD/EDDN@4ad669bb7bbe1eae080e4c354e786dca4db91f35`

## Validation

- Release x64 application build: **0 warnings, 0 errors**
- Production-referenced EDDN suite: **74 passed**
- `git diff --check`: clean
- Synthetic merge with PR #1025 at `8c36612d`: application build **0 warnings, 0 errors**; **74 EDDN tests** and **62 Inara tests** passed

The test project references the production application assembly through `InternalsVisibleTo`, is included in the solution, and runs explicitly in CI.


Zaaeen:
This is a really cool feature I will be enabling the moment it comes live..

SrvSurvey/Main.cs
```diff
@@ -38,10 +39,11 @@ internal partial class Main : FixedForm
...
         public KeyboardHook hook;
 
-        public Main()
-        {
-            Game.log("Starting main form...");
-            form = this;
+        public Main()
+        {
+            Game.log("Starting main form...");
+            eddn = new EDDN();
```
njthomson: *(I've yet to review much of this PR but...)*
I wonder if it would make sense for `EDDN` to be an instance member of `Game` ... similar to the `Inara` class?
I would not make `Main` be the owner of the instance, especially if we are passing it into `Game` via the ctor.

SrvSurvey/Main.cs
```diff
@@ -315,9 +317,10 @@ private void timer1_Tick(object sender, EventArgs e)
             {
                 try
                 {
-                    lastProcCheck = DateTime.Now;
-                    var procs = Elite.GetGameProcs();
-                    Elite.hadManyGameProcs = procs.Length > 1;
+                    lastProcCheck = DateTime.Now;
+                    Elite.refreshManyGameProcs();
+                    if (Game.settings.eddnUploadEnabled)
```
njthomson: *(nit)* I would position this after line 326 - so that the heaviest operation is last.

SrvSurvey/Main.cs
```diff
@@ -315,9 +317,10 @@ private void timer1_Tick(object sender, EventArgs e)
...
+                        eddn.refreshRuntimeSafety();
```
njthomson: `Elite.refreshManyGameProcs();` gets called again from within `eddn.refreshRuntimeSafety();` ... we don't want to be doing this twice in quick succession.

SrvSurvey/game/Game.cs
```diff
@@ -613,12 +618,13 @@ private void initializeFromJournal(LoadGame? loadEntry = null)
...
                 log($"Game.initializeFromJournal: USING {this.Commander} (FID:{this.fid}), journals.Count:{journals?.Count}");
 
-                // set EDDN upload header
-                if (journals?.Entries.Count > 0)
-                {
-                    var gameFileHeader = (Fileheader)journals.Entries.First();
-                    EDDN.header = new UploadPayloadHeader(loadEntry.Commander, gameFileHeader.gameversion, gameFileHeader.build);
-                }
+                // set EDDN upload header
+                var gameFileHeader = journals?.Entries.OfType<Fileheader>().FirstOrDefault();
```
njthomson: Similar to the Inara PR ... I'm thinking we should keep a direct reference to the `FileHeader` from `JournalFile`

SrvSurvey/game/Game.cs
```diff
@@ -112,7 +111,8 @@ private static void removeExcessLogFiles()
         public static EDSM edsm { get; private set; }
         public static Git git { get; private set; }
         public static RavenColonial.RavenColonial rcc { get; private set; }
-        public static EDDN eddn { get; private set; }
+        private readonly EDDN eddnService;
```
njthomson: The original design was for `EDDN` to be a static singleton ... I think making it an instance member of `Game` is a good move ... but I'm curious to see why we would need both `EDDN` and `EddnSessionPublisher`. Before I review that code, it feels like we should be able to contain what ever complexity is needed within a single object.

tests/SrvSurvey.EddnTests/SrvSurvey.EddnTests.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0-windows</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
+    <PlatformTarget>x64</PlatformTarget>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
njthomson, line 12: Add the following to avoid warnings in Visual Studio:
``` xml
<Platforms>x64</Platforms>
<Configurations>Debug;Release;ReleaseOnce;DebugNoPS</Configurations>
```

SrvSurvey/game/JournalExpansionFlags.cs
```diff
@@ -0,0 +1,24 @@
+using Newtonsoft.Json.Linq;
+
+namespace SrvSurvey.game
+{
+    internal readonly record struct JournalExpansionFlags(
```
njthomson: This is over engineered and redundant. We should just keep a reference to the `LoadGame` entry on `JournalFile` and use that.

I have something specific in mind for `FileHeader` ... let me make that change quickly...

SrvSurvey/Main.cs
Fenris159: Mentioned in discord, basically you don't want to flush the data with the loss of game, shutdown, or power event. Exploration data may not update due to whatever so the data is batched and retried later which is why it lives in Main.

SrvSurvey/game/Game.cs
Fenris159: Your right. I thought I aligned this with Inara after those lessons but looks like I missed this.

SrvSurvey/Main.cs
Fenris159: Strange I thought I caught this before. I'll get on it.

tests/SrvSurvey.EddnTests/SrvSurvey.EddnTests.csproj
Fenris159: That came from a platform config property that was misconfigured.

SrvSurvey/game/Game.cs
Fenris159: This worked for Inara because nothing Inara cares about is allowed to outlive `Game`. If you collapse them into `Game` you get a static singleton bug. Delivery loop that still wants to run after `Game.Dispose`, so it could either hold a dead `Game` or start reading `Game.activeGame`. 

Collapse them onto `Main` and session state "signal batch, companion delay, commander header" sits on an object that survives commander switches. `EddnSessionPublisher `is going to behave like Inara however `EDDN` has to be separate because of it's durable outbox.

njthomson: I'll have to dig into those classes to have a proper opinion then.

SrvSurvey/Main.cs
njthomson: In that case, I would recommend a static singleton hanging off of `Game`.
*(Using `Main` works, but I prefer to keep that tied to visual things, and stateful journal things tied to `Game`)*

Fenris159:
- Accepted current upstream `main`, including Inara and cached `Fileheader`/`LoadGame` changes.
- Moved the application-lifetime EDDN singleton from `Main` to `Game`, initialized after settings load and disposed at application shutdown.
- Kept `EddnSessionPublisher` session-scoped, so durable delivery survives `Game.Dispose`.
- Replaced `JournalExpansionFlags` with nullable fields on cached `LoadGame`.
- Added the requested x64/configuration properties and corrected solution mappings.

SrvSurvey/Program.cs
```diff
@@ -123,8 +123,19 @@ static void Main(string[] args)
...
-                Application.Run(mainForm);
+                try
+                {
+                    if (!invokePostProcessor)
+                        Game.initializeEddn();
+
+                    Form mainForm = invokePostProcessor ? new FormPostProcess(args.LastOrDefault()) : new Main();
+                    Application.Run(mainForm);
+                }
+                finally
```
njthomson, line 134: I'm not sure this is the best place for this. If we're here the process is about to die - I'm wondering if there would be enough time for any pending network operations to complete? And you need that IF for post-processing.

If you relocate this to Main.cs - to OnClosing, that would give you an opportunity to cancel the closing until you know network operations complete, then call `.close()` on the main form again.

SrvSurvey/game/Game.cs
```diff
@@ -112,8 +122,11 @@ private static void removeExcessLogFiles()
         public static EDSM edsm { get; private set; }
         public static Git git { get; private set; }
         public static RavenColonial.RavenColonial rcc { get; private set; }
-        public static EDDN eddn { get; private set; }
-        internal Inara? inara;
+        private static EDDN? eddnService;
```
njthomson, line 125: I think having `initializeEddn` is overkill. Why not just keep it as it was before?
As a static constructor, it will get called as soon as any external code references the type `Game` long before any instances and ahead of the main window getting created.

This removes the need to do null checking anywhere else too.

SrvSurvey/game/Game.cs
```diff
@@ -789,6 +809,21 @@ private void initializeFromJournal(LoadGame? loadEntry = null)
                 this.cargoFile.Inventory.Clear();
             }
 
+            this.eddnSession?.Dispose();
+            this.eddnSession = eddnHeader == null
```
njthomson, line 813: (nit) This feels more readable...

``` C#
if (eddnHeader != null)
  this.eddnSession = new EddnSessionPublisher(...);
```

SrvSurvey/game/Game.cs
```diff
@@ -1116,13 +1151,34 @@ private void Journals_onJournalEntry(IJournalEntry? entry, JObject raw)
...
+                            this.systemBody?.id,
+                            this.systemBody?.type is SystemBodyType.Giant
+                                or SystemBodyType.SolidBody
+                                or SystemBodyType.LandableBody
+                                ? "Planet"
+                                : null));
+                }
+                catch (Exception ex)
+                {
+                    Game.log($"EDDN ignored an exception processing '{eventName}' so other journal consumers can continue:\r\n{ex}");
```
njthomson, line 1177: We want visibility into any such failures, it is good to suppress then from users but you & I should know about them. I recommend adding `Debugger.Break();` after logging to get our attention.

SrvSurvey/forms/FormEddnIntegration.cs
```diff
@@ -0,0 +1,180 @@
...
+                    MessageBoxIcon.Warning) != DialogResult.Yes)
+            {
+                return;
+            }
+
+            var previous = Game.settings.eddnUploadEnabled;
+            try
+            {
+                Game.settings.eddnUploadEnabled = enabled;
+                Game.settings.Save();
```
njthomson, line 139: Alas, we cannot have this here. FormSettings controls when global settings are saved, or not if the user hits Cancel.

SrvSurvey/forms/FormEddnIntegration.cs
```diff
@@ -0,0 +1,180 @@
...
+                eddn.setEnabled(enabled);
+                DialogResult = DialogResult.OK;
+                Close();
+            }
+            catch (Exception ex)
+            {
+                Game.settings.eddnUploadEnabled = previous;
+                List<string> rollbackErrors = [];
+                try
```
njthomson, line 148: This is rather* over-engineered. What kind of exceptions/failures do you expect? How would you expect things to be that much different when you attempt saving just below?

Unless there's a class of errors you anticipate with any frequency, I'd just route users to the error-submit form and let them tell us if there's a real problem.

*(** *this is British for "this is way way over engineered)*

SrvSurvey/game/JournalEntry.cs
```diff
@@ -66,8 +66,8 @@ public class LoadGame : JournalEntry
 
         public string FID { get; set; }
         public string Commander { get; set; }
-        public bool Horizons { get; set; }
-        public bool Odyssey { get; set; }
+        public bool? Horizons { get; set; }
```
njthomson, line 69: Did you observe any error with these being not-nullable?

When these are missing, like in really old journal files, they would default to `false` - which would be the correct state, no?

SrvSurvey/game/JournalFile.cs
```diff
@@ -38,10 +38,10 @@ static JournalFile()
         public string? cmdrFid { get; private set; }
         /// <summary> Returns true if this is from a Legacy journal, or false if this from a Live/Odyssey journal. Assumes NOT legacy if we don't have the Fileheader line yet. </summary>
         public bool isLegacy => this.fileheader?.Odyssey == false;
-        /// <summary> Returns true the player paid for the Odyssey DLC </summary>
-        public bool isGameOdyssey => this.lastLoadGame?.Odyssey ?? false;
-        /// <summary> Returns true the player has the Horizons expension (which I think everyone does now) </summary>
-        public bool isGameHorizons => this.lastLoadGame?.Horizons ?? false;
+        /// <summary> Returns whether the player paid for the Odyssey DLC, or null when the journal omitted the flag. </summary>
+        public bool? isGameOdyssey => this.lastLoadGame?.Odyssey;
```
njthomson, line 42: I think we do ourselves a favour by returning a non-nullable value. It only places a burden on calling code to rationalize between three states rather than two.
*(Sometimes that can a good thing, don't get me wrong, but in this specific case, I think there's no benefit)*

SrvSurvey/Elite.cs
```diff
@@ -54,6 +54,26 @@ public static Process[] GetGameProcs()
             return Process.GetProcessesByName("EliteDangerous64");
```
njthomson, line 54: *(I like what you have at line 72)*
Can you add a TODO comment here that we should be disposing these Process objects in calling code.

SrvSurvey/Main.cs
```diff
@@ -326,9 +325,18 @@ private void timer1_Tick(object sender, EventArgs e)
...
+
+                    if (Game.settings.eddnUploadEnabled)
+                        Game.eddn.refreshRuntimeSafety(Elite.hadManyGameProcs);
                 }
                 catch (Exception ex)
                 {
+                    // EDDN must fail closed when shared file attribution cannot
+                    // be established. Other game/process handling remains live.
+                    if (Game.settings.eddnUploadEnabled)
+                        Game.eddn.setSuspended(
```
njthomson, line 337: Would it make sense for `EDDN.refreshRuntimeSafety` to have an internal try/catch and make this call internally?

SrvSurvey/net/EddnConsentFile.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace SrvSurvey.net
+{
+    /// <summary>
+    /// Reads the application-wide EDDN consent value without mutating the
+    /// process-local Settings instance. Multiple SrvSurvey processes share the
+    /// same settings file, so an opt-out in any instance must be observed by
+    /// whichever process currently owns the durable outbox.
+    /// </summary>
+    internal static class EddnConsentFile
```
njthomson, line 12: Whilst rather impressive, I think this is overkill. No other setting in SrvSurvey has this cross-process ability. I'd like to think users understand that settings do not apply atomically across processes *(though you never know for sure)*. Still, I'd be shocked if anyone ever runs afoul of this.

However ... I recommend people use the `runas` Windows command now for multi-boxing *(as it is so so much more reliable)* and what you have here isn't going to help in those cases *(as those have a completely different settings.json file in a different folder)*.

So ... rather than have to explain to people why honouring settings across processes may work sometimes, and get into the details of why it would or would not ... I think it safe enough drop this cross-process ability.

SrvSurvey/net/EDDN.cs
```diff
@@ -1,452 +1,235 @@
...
-            upload(message, "https://eddn.edcd.io/schemas/journal/1").justDoIt();
+            var queued = transport.prepare(
+                prepared.message,
+                prepared.schemaRef,
+                header);
+            return outbox.enqueue(queued);
         }
 
-        public void onJournalEntry(Game game, Scan _, JObject raw)
+        private void refreshGameProcessSafety(bool? hasMultipleEliteProcesses)
```
njthomson, line 155: Make this:
``` C#
private void refreshGameProcessSafety(bool hasMultipleEliteProcesses)
```
... the `bool?` is not helpful.

You have only 1 caller currently. Rather than allowing extra complexity by callers using `null` *(it isn't even optional)* ... just enforce they provide a boolean value and remove the optional call to `?? Elite.refreshManyGameProcs()`

SrvSurvey/net/EddnOutbox.cs
```diff
@@ -0,0 +1,870 @@
+using Newtonsoft.Json;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace SrvSurvey.net
+{
+    /// <summary>
+    /// A durable, ordered EDDN replay queue modelled after EDMC's sender queue.
```
njthomson, line 8: Maybe add a GitHub link to the EDMC code?

SrvSurvey/net/EddnOutbox.cs
```diff
@@ -0,0 +1,870 @@
...
+        private bool enabled;
+        private bool suspended;
+        private bool loadingTruncated;
+        private bool ownershipWarningReported;
+        private volatile bool disposed;
+
+        internal EddnOutbox(
+            string filepath,
+            EddnTransport transport,
+            Action<string>? log = null,
```
njthomson, line 50: What's wrong with using `Game.log(..)` like everywhere else? 
*(Other things that are static singletons hanging off of `Game` use it in a lot of other places)*

I don't like changing the logging pattern, unless there's a strong reason for it.

SrvSurvey/net/EddnOutbox.cs
```diff
@@ -0,0 +1,870 @@
...
+
+                return true;
+            }
+            catch (Exception exception) when (
+                exception is IOException or UnauthorizedAccessException)
+            {
+                if (!ownershipWarningReported)
+                {
+                    messages.Add(
+                        "EDDN uploads are paused because another SrvSurvey instance owns the local outbox.");
```
njthomson, line 774: Under what conditions might this happen? Is it logic pulled over from EDMC code?

The mechanism of acquiring the outbox feels like overkill as I don't allow SrvSurvey to run twice for the same account.

SrvSurvey/net/EddnOutbox.cs
```diff
@@ -0,0 +1,870 @@
...
+
+                loaded.Add(candidate.item);
+                persistedBytes[candidate.item.id] = candidate.length;
+                storeBytes += candidate.length;
+            }
+
+            return loaded;
+        }
+
+        private void migrateLegacyStore(
```
njthomson, line 581: This is all new code for us, what legacy store would we have to worry about?

Or is it Legacy game mode? I thought we were going to ignore those?

SrvSurvey/net/EddnOutbox.cs
```diff
@@ -0,0 +1,870 @@
...
+            if (item == null) return;
+            if (string.IsNullOrWhiteSpace(item.schemaRef)) return;
+            item.schemaRef = EddnTransport.applySchemaPolicy(
+                item.schemaRef,
+                EddnTransport.testSchemasEnabled);
+        }
+
+        private static void quarantine(string path, List<string> messages)
+        {
+            var backup = path
```
njthomson, line 709: May add a `Debugger.Break();` in here? Is this a truly exceptional case? Should you or I know about it if it starts happening?

SrvSurvey/net/EddnOutbox.cs
njthomson: Reviewing more code ... I think we should keep function `writeLogs` that is good to do outside of the many `lock` blocks.

SrvSurvey/net/EddnOutbox.cs
```diff
@@ -0,0 +1,870 @@
...
+        private readonly string storeFolder;
+        private readonly string ownershipPath;
+        private readonly EddnTransport transport;
+        private readonly Action<string> log;
+        private readonly Func<DateTimeOffset> utcNow;
+        private readonly bool automaticProcessing;
+        private readonly Func<bool> runtimeUploadAllowed;
+        private readonly int maximumPendingMessages;
+        private readonly long maximumStoreBytes;
+        private readonly object sync = new();
```
njthomson, line 31: Does all the locking happen off of the main/UI thread? I have SrvSurvey use the game process as parent, hence locking our own process will mean the game can stop receiving input too. *(I've crashed my ship a few times because of this)*

And same commend for EDDN.cs too.

njthomson (commented):
Added a number of comments for you ... I'll be using this build locally when next I play - will let you know how I get on.

SrvSurvey/Program.cs
Fenris159: The pending messages are already saved to disk before upload, so shutdown doesn’t need to wait for network requests to finish and they’ll retry next time SrvSurvey starts. I’d prefer to keep disposal in `Program`’s `finally`, since it also covers non-standard exits and already excludes post-processing.

SrvSurvey/game/Game.cs
Fenris159: Good point about when the static constructor runs. That’s actually why I avoided creating EDDN there. `Game `can be referenced while settings are still loading, but EDDN needs those settings immediately. The explicit initializer ensures EDDN is only created after `Settings.Load()` completes.

SrvSurvey/game/Game.cs
Fenris159: What about this? It preserves the important null assignment while giving your preferred `if `structure: 

```
this.eddnSession?.Dispose();
this.eddnSession = null;

if (eddnHeader != null)
{
    this.eddnSession = new EddnSessionPublisher(
        Game.eddn,
        eddnHeader,
        Game.settings.watchedJournalFolder,
        this.systemData is { address: > 0, starPos.Length: 3 }
            ? new EddnLocationContext(
                this.systemData.name,
                this.systemData.address,
                this.systemData.starPos.ToArray())
            : null,
        Game.log);
}
```

SrvSurvey/game/Game.cs
Fenris159: I can use `if (Debugger.IsAttached) Debugger.Break();` after logging so it's not unconditional. I wouldn't want to invoke just in time debugging.

SrvSurvey/forms/FormEddnIntegration.cs
Fenris159: I'll restage this value in `FormSettings` and only persist and apply it when the parent save succeeds.

SrvSurvey/forms/FormEddnIntegration.cs
Fenris159: You’re right that the rollback logic is over-engineered. Looking at it more closely, it also doesn’t provide the protection I intended...
I don’t think the right correction is just deleting the error handling, because EDDN consent should only become active after we know the setting was saved successfully.

My proposed approach is:
1. Let the EDDN dialog return the user’s selected value without saving or changing the running service.
2. Keep that value staged in `FormSettings`, so cancelling the main Settings form discards it.
3. Add a small synchronous settings-save path that either succeeds or reports/throws the write failure.
4. Only call `eddn.setEnabled(...)` after that save succeeds.
5. If saving fails, restore the in-memory value and show the normal error-report form.

That should let us remove the nested rollback code while preserving the important guarantee that the persisted consent setting and the active uploader cannot drift apart.

SrvSurvey/game/JournalEntry.cs
Fenris159: Missing means unknown; `false `explicitly means the Commander does not have that expansion. EDDN describes these fields as copied from `LoadGame` “when present,” and they are optional in the schema. Defaulting missing data to false would publish an unsupported claim.
[EDDN schema evidence](https://github.com/EDCD/EDDN/blob/153e62426eb58b6cd23def535cb4b7e1082e777f/schemas/fcmaterials_journal-v1.0.json)

SrvSurvey/game/JournalFile.cs
Fenris159: The nullable value does add a little work for callers, but these properties are currently used for EDDN message construction, where the distinction is useful: send `true `or `false `when the journal provides it, and omit the field when it does not.
For that reason, I recommend keeping these two properties nullable.

SrvSurvey/Elite.cs
Fenris159: I'll write that in and I can plan a dedicated PR for cleaning up the older callers throughout the codebase.

SrvSurvey/Main.cs
Fenris159: That makes sense from an encapsulation standpoint, but `Main `already performs the process check because it also needs the result for its UI. Moving the check into EDDN would duplicate that work.
I’d keep passing the result into EDDN, but simplify the failure handling so EDDN remains responsible for putting itself into a safe, suspended state.

SrvSurvey/net/EddnConsentFile.cs
Fenris159: Agreed. Looking at it again, the cross-process consent handling protects a very narrow edge case and adds more complexity than it is worth.
I’ll remove EddnConsentFile and the related warning state, tests, and repeated settings-file reads. Each running SrvSurvey process will use the EDDN setting it loaded normally, consistent with the rest of the application.
Disabling EDDN will still immediately stop delivery and clear pending uploads for that process. I’ll retain the exclusive outbox lease because it separately prevents multiple processes from modifying or delivering the same queue, as well as the existing pause while multiple Elite clients are running.

SrvSurvey/net/EDDN.cs
Fenris159: Agreed. The nullable Boolean was being used as a hidden instruction: a value meant “use this result,” while `null `meant “perform the process check yourself.” That is not especially clear.
I’ll replace it with two overloads: one that performs a fresh process check and another that accepts the Boolean already calculated by `Main`. The internal helper will then always require a non-nullable Boolean. This preserves both existing call paths without giving `null` a special meaning.

SrvSurvey/net/EddnOutbox.cs
Fenris159: That’s fair. There is no need for the caller to choose the logging implementation, so I can make the normal `EddnOutbox `constructor use `Game.log` directly, matching the rest of SrvSurvey.
I’ll keep a separate internal constructor overload that accepts a logger for the test project. That lets the tests capture diagnostics without initializing or interacting with the UI logger.
I can keep `writeLog `and `writeLogs ` so logging still happens outside the queue `lock`s. This should keep the normal application pattern while preserving the small test seam we need.

SrvSurvey/net/EddnOutbox.cs
Fenris159: I checked this against the process-startup code, and I think we should keep the outbox ownership lease.
SrvSurvey normally prevents a second instance, but `FormStartNewCmdr `deliberately starts additional instances using `-fid`, which bypasses that check. When those instances run under the same Windows account, they share the same data folder and EDDN queue. Without the lease, two processes could load, upload, or delete the same queue entries concurrently.
`runas` instances use separate data folders, so the lease does not affect those.
I did find one improvement: if an instance cannot acquire the lease, it does not automatically retry after the owner exits. I recommend retaining the lease while adding a quiet, delayed retry so ownership can transfer automatically.

SrvSurvey/net/EddnOutbox.cs
Fenris159: Good catch! I was using AI to help me handle this process and it assumed I had users running earlier builds before I changed the Queue Format so it wrote this safeguard to handle old builds cleanly. It's totally redundant and I missed it!

SrvSurvey/net/EddnOutbox.cs
Fenris159: This path handles damaged, incomplete, locked, or manually modified queue files. Those are recoverable environmental problems rather than necessarily programming errors.
The code already preserves the affected file as a `.bad-*` file and records the reason in the logs, which gives us the evidence needed to investigate it. I would avoid an unconditional `Debugger.Break()` because it could interrupt a normal release build.
I can add `if (Debugger.IsAttached) Debugger.Break();` if we want extra visibility during development, but I think quarantine and logging are sufficient here.

SrvSurvey/net/EddnOutbox.cs
Fenris159: I see your concern. The network upload already runs in the background, and the locks in EDDN.cs only protect short state changes. However, journal callbacks run on the UI thread, and the enqueue path currently performs synchronous queue-file operations while holding its serialization locks.
I’ll move durable persistence to one application-owned background worker. The journal handler will give it an immutable message, and the worker will persist messages serially before marking the enqueue successful. This preserves ordering and queue safety without making the UI wait for disk access.
I’ll retain the locks needed to protect shared state. I can move blocking file work off the UI thread by doing that instead of simply removing synchronization.

Fenris159: This is probably the highest priority item I'll have to address.

PR #1028, opened 2026-07-30
Fix malformed Bacterium Informem reference criterion

Fenris159:
## Summary

- fix the misplaced closing bracket in the Bacterium Informem atmosphere-composition criterion
- regenerate `data/bio-criteria.zip` from the corrected source directory
- bump `bioCriteria` in both update manifests so installed clients fetch the correction

## Why

The malformed clause is valid JSON but invalid in the biology criteria grammar:

```text
atmosComp [Nitrogen >= 100] | SulphurDioxide >= 28 | Argon >= 0.06]
```

The extra bracket closes the composition after its first value, leaving the remaining alternatives outside the clause. Keeping all three alternatives inside one bracket restores the intended OR composition and prevents the downloaded biology catalog from failing to load.

## Reference-data audit

- parsed all 773 tracked JSON files and every JSON member of the three published archives (1,524 documents total), including duplicate-key checks
- validated all 21 biology files, 897 nodes, and 2,842 query clauses against the criteria grammar
- checked every `bio-criteria.zip`, `guardian.zip`, and `settlements.zip` member byte-for-byte against its source directory
- verified the source and GitHub Pages update manifests remain synchronized

No other malformed reference entries were found.

## Validation

- `dotnet build SrvSurvey/SrvSurvey.csproj --configuration Release --verbosity minimal`
- complete reference-data syntax, duplicate-key, criteria-grammar, and archive-consistency audit


njthomson (approved):
Nice catch

PR #1035, opened 2026-08-09
Automatically remove sold exploration data from exploration estimates

Fenris159:
## Summary

- Persist estimated exploration rewards by star system while retaining the existing aggregate shown on the main window.
- Consume `SellExplorationData` and `MultiSellExplorationData` after the completed journal event has passed through normal local processing.
- Remove only matched sold systems from the unsold estimate, page by page, and refresh the main-window value immediately.
- Keep sale handling idempotent and clear the per-system ledger through the existing manual exploration reset.

## Why

The exploration trip counter previously accumulated estimated scan value but could not identify which systems contributed to it. Universal Cartographics sells data page by page, so the local estimate remained unchanged after individual systems were sold.

This change records each system's contribution and uses the system names supplied by Frontier's completed sale events to subtract only the sold data.

## Behavior and compatibility

- Historical per-system scan files are not deleted or modified.
- Selling data changes only the estimated reward total; jump, distance, scan, DSS, and landing trip statistics remain intact.
- Duplicate or replayed sale events cannot deduct the same tracked system twice.
- Existing aggregate rewards recorded before this change have no system attribution. Automatic deductions apply to rewards tracked after this version or after the commander next uses the existing manual reset.

## Checks

- `git -c core.whitespace=cr-at-eol diff upstream/main...origin/exploration-sold-data-tracking --check` passed.
- `dotnet build SrvSurvey/SrvSurvey.csproj` succeeded with 0 errors and the repository's existing warnings.

### DIFF
--- a/SrvSurvey/game/CommanderSettings.cs
+++ b/SrvSurvey/game/CommanderSettings.cs
@@ -160,8 +160,6 @@ namespace SrvSurvey.game
         public long organicRewards;
 
         public long explRewards;
-        [JsonProperty(NullValueHandling = NullValueHandling.Ignore, DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public Dictionary<string, long>? explRewardsBySystem;
         public double distanceTravelled;
         public int countJumps;
         public int countScans;
@@ -203,6 +201,9 @@ namespace SrvSurvey.game
         [JsonProperty(NullValueHandling = NullValueHandling.Ignore, DefaultValueHandling = DefaultValueHandling.Ignore)]
         public int countRadicoidaUnica;
 
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore, DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public Dictionary<string, long>? explRewardsBySystem;
+
         #endregion
 
         public void applyExplReward(long reward, string reason, string systemName)
@@ -214,7 +215,7 @@ namespace SrvSurvey.game
             {
                 this.explRewardsBySystem ??= new Dictionary<string, long>(StringComparer.OrdinalIgnoreCase);
                 var trackedName = this.explRewardsBySystem.Keys.FirstOrDefault(name =>
-                    name.Equals(systemName, StringComparison.OrdinalIgnoreCase)) ?? systemName;
+                    name.like(systemName)) ?? systemName;
                 this.explRewardsBySystem[trackedName] = this.explRewardsBySystem.GetValueOrDefault(trackedName) + reward;
             }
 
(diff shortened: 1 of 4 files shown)


SrvSurvey/game/CommanderSettings.cs
```diff
@@ -203,13 +205,61 @@ private static void seedCommanderSettings()
...
+        public void applyExplReward(long reward, string reason, string systemName)
         {
             Game.log($"Gained: +{reward.ToString("N0")} for {reason}");
             this.explRewards += reward;
 
+            if (!string.IsNullOrWhiteSpace(systemName))
+            {
+                this.explRewardsBySystem ??= new Dictionary<string, long>(StringComparer.OrdinalIgnoreCase);
+                var trackedName = this.explRewardsBySystem.Keys.FirstOrDefault(name =>
+                    name.Equals(systemName, StringComparison.OrdinalIgnoreCase)) ?? systemName;
```
njthomson: > name.Equals(systemName, ...

Try `name.like(systemName)` - I have an extension method for this.

SrvSurvey/game/CommanderSettings.cs
```diff
@@ -160,6 +160,8 @@ private static void seedCommanderSettings()
         public long organicRewards;
 
         public long explRewards;
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore, DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public Dictionary<string, long>? explRewardsBySystem;
```
njthomson: This is likely to become a long list, can you move it down to be above `#endregion` - so that it is last in the .json file.

SrvSurvey/game/CommanderSettings.cs
```diff
@@ -203,13 +205,61 @@ private static void seedCommanderSettings()
...
+
+            var deductedRewards = Math.Min(this.explRewards, removedRewards);
+            this.explRewards -= deductedRewards;
+            if (this.explRewardsBySystem.Count == 0)
+                this.explRewardsBySystem = null;
+
+            Game.log($"{eventName}: removed {deductedRewards.ToString("N0")} estimated exploration credits for {matchedSystems} sold system(s).");
             PlotBase2.renderAll(Game.activeGame);
             this.Save();
+            Game.activeGame?.fireUpdate(true);
```
njthomson, line 262: I think if we're forcing an update, there's no need for `PlotBase2.renderAll(..)`

SrvSurvey/game/Game.cs
```diff
@@ -1110,7 +1110,22 @@ private void Journals_onJournalEntry(IJournalEntry? entry, JObject raw)
                         eddn.onJournalEntry(this, (dynamic)entry, raw);
                 }
 
-                // finally, let active quests know about this
+                // Frontier emits these only after a Cartographics page has finished selling.
+                // Reconcile the local estimate independently of optional upload integrations.
+                if (entry is SellExplorationData sellExplorationData)
```
njthomson: This is the wrong place for this. You could add a method below `private void onJournalEntry(SellExplorationData entry)` with your logic from here.

... and another method for `MultiSellExplorationData`

Fenris159:
I saw the journal paths  you already had in place and moved the process into the existing pipeline.

njthomson (approved):
Nice work.

PR #1053, opened 2026-09-01
Add Guardian map marker alignment offsets

Fenris159:
## Summary
- port the Guardian map-marker alignment offset from Avalonia PR Fenris159/SrvSurvey#103 into legacy SrvSurvey
- add a minimal **Guardians > Correct Map Alignment Here** command that uses the commander's current latitude/longitude as the corrected survey origin
- apply the saved offset to Guardian map POIs, group labels, component markers, hit-testing, and the live surface overlay
- read and write Avalonia-compatible top-level `mapMarkerOffset` metadata while preserving optional catalogue metadata during legacy round-trips
- retain the physical correction when the site's heading is edited

## Validation
- `dotnet test SrvSurvey.Tests/SrvSurvey.Tests.csproj --no-restore` (69 passed)
- Avalonia Guardian offset/store compatibility tests (8 passed)
- `dotnet build SrvSurvey/SrvSurvey.csproj --no-restore -c Release` (0 errors; existing nullable warnings remain)
- `dotnet format SrvSurvey.sln --no-restore --verify-no-changes --include ...`
- `git -c core.whitespace=cr-at-eol diff --check`

Fork review: Fenris159/SrvSurvey#104. Both PRs use the identical commit and branch.

### DIFF
--- a/SrvSurvey.Tests/GuardianMapMarkerOffsetTests.cs
+++ b/SrvSurvey.Tests/GuardianMapMarkerOffsetTests.cs
@@ -1,0 +1,153 @@
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using SrvSurvey.game;
+using SrvSurvey.units;
+using Xunit;
+
+namespace SrvSurvey.Tests;
+
+public sealed class GuardianMapMarkerOffsetTests
+{
+    [Fact]
+    public void CalculatesSamePortableMapOffsetAsAvalonia()
+    {
+        var offset = GuardianMapMarkerOffsetCalculator.calculate(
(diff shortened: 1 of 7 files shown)


Fenris159:
`mapMarkerOffset` top-level metadata is the format I used in survey files that would allow surveys to maintain readability between both program versions and act like a special configuration for all the following data, so if you decide to adopt this code in any way my only request would be to either adopt that format or otherwise similar enough I can make minor adjustment to re-adopt back for maximum compatibility.

Fenris159:
I left this here for reference but let me know if I can delete it or if you want the code to get merged later.

